Issue #101, opened 2015-10-13
Doesn't work in IE9 as advertised


stoikerty:
Hey Matt,

nice job with this project, looks very promising! The narrow focus makes it useable for React which is exactly what I was looking for (a physics engine that I can incorporate into `state`) :)

Now onto business... I hate to be the guy pointing this out. You advertise popmotion as being compatible with IE9. Your main site [popmotion.io](http://popmotion.io) throws this error:
```
SCRIPT438: Object doesn't support property or method 'now' 
site.js, line 1257 character 2
```

I haven't looked into it in depth, it feels like this might have to do with how `performance.now` is implemented.

InventingWithMonster:
Thanks for pointing this out, that shouldn't be happening. 

Let us know how you get on with React. I'd like to provide a more specific implementation for it.

InventingWithMonster:
3.3.3

stoikerty:
that was quick! :+1: 

As a sidenote, maybe [the repo for `popmotion.io`](https://github.com/InventingWithMonster/popmotion-frontend) could live in popmotion's github-profile when you feel that it's ready?

InventingWithMonster:
Yeah, it's going to, though like you say "when it's ready" - it uses an old
version of Popmotion with old syntax, not strictly something I want people
to learn from just yet.

InventingWithMonster:
@stoikerty https://github.com/Popmotion/popmotion-frontend

PR #103, opened 2015-10-13
fix IE9 `performance.now`-issue

stoikerty:
see #101

### DIFF
--- a/lib/inc/utils.js
+++ b/lib/inc/utils.js
@@ -251,7 +251,7 @@ module.exports = {
         @return [timestamp]: Current UNIX timestamp
     */
     currentTime: function currentTime() {
-        return typeof performance !== "undefined" ? performance.now() : new Date().getTime();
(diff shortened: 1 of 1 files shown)


lib/inc/utils.js
```diff
@@ -251,7 +251,7 @@ module.exports = {
         @return [timestamp]: Current UNIX timestamp
     */
     currentTime: function currentTime() {
-        return typeof performance !== "undefined" ? performance.now() : new Date().getTime();
+        return (performance && performance.now) ? performance.now() : new Date().getTime();
```
InventingWithMonster: @stoikerty Hey man my bad for not noticing this but all changes happen in `/src` - I've just been running round the codebase wondering why your fix had reverted! This'll be in properly with `3.6.0`

stoikerty: nice one for noticing this :ok_hand:

PR #158, opened 2016-06-06
Prefixer not adding opt for dash case on Transforms.

jamieowen:
Hey man,

Just managed to update to the latest version containing the css prefixer/dash case additions..
Except just noticed that the transforms were not being covered..

Cheers,

Jamie


InventingWithMonster:
Nice catch, thanks for this!

Issue #162, opened 2016-06-18
Unit types are not respected


null-null-null:
If I provide a unit type other than `px`, it will simply be converted into `px`. For instance, I could try something like this:

``` js
const myTween = motion.tween({
    values: {
        x: '200%'
    }
});

myTween.on(document.getElementById('ball')).start();
```

And it would result in this:

`<div id="ball" style="transform: translateX(200px) translateZ(0px);"></div>`

[Here](http://codepen.io/nullnullnull/pen/wWzoJp?editors=0010) is a codepen demonstrating the issue.

I'm still new to the popmotion codebase, but spent a bit of time trying to debug the issue. I wasn't able to find the the root cause, but it looks like there are possibly two issues. The first is with [this line](https://github.com/Popmotion/popmotion/blob/master/src/value-types/unit.js#L12), which I believe would only run if the unit was `NaN`.

The second thing I found was that when running this:

``` js
const myTween = motion.tween({
    values: {
        x: '200%'
    }
});
```

The actual unparsed `200%` gets passed into [parse](https://github.com/Popmotion/popmotion/blob/master/src/value-types/unit.js#L9). However, when running this:

```
myTween.on(document.getElementById('ball')).start();
```

The `unparsed` comes in as `200`, which is why it uses the default `px` unit. Hope that helps!

InventingWithMonster:
I'm just looking into this now - I think a problem is Popmotion already "knows" this type of value is in pixels, when of course there's a few different types it could potentially be. I'm thinking a potential approach could be to define permitted value types along with just one default type.

null-null-null:
Let me know if there's anything I can do to help! My main observation was that Popmotion was "forgetting" the type [here](https://github.com/Popmotion/popmotion/blob/master/src/actions/Flow.js#L42). I'm curious how it remembers the value, though. Perhaps we could store the type in a similar manner?

PR #163, opened 2016-06-20
Inherit actions so unit is respected

null-null-null:
This is an attempt to resolve https://github.com/Popmotion/popmotion/issues/162: essentially, units are lost and always return to their value type default.

This PR fixes the problem. That being said, although I've got this working locally and all tests still pass, I'm not familiar enough with the codebase to have confidence that I'm not breaking something. If nothing else, hopefully this PR will help you find the source of the problem.

### DIFF
--- a/src/actions/Flow.js
+++ b/src/actions/Flow.js
@@ -39,7 +39,7 @@ class Flow extends Action {
     // Create values on actor that don't exist
     for (let key in inheritedAction.values) {
       if (inheritedAction.values.hasOwnProperty(key) && !this.values.hasOwnProperty(key)) {
-        newValues[key] = {};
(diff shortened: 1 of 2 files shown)


villavicencio:
@InventingWithMonster is this something that could get merged down?

InventingWithMonster:
Sorry I neglected this for so long - closing the PR as I rewrote this to the point of killing most of these annoying bugs. They all stemmed out of Popmotion being too opinionated. Lesson learned.

villavicencio:
Hey. No problem, I actually updated to 6.3.5 a couple days ago. Took a couple hours to rewrite some code to fit in with the new API but everything is working great. Great job on the update.

PR #176, opened 2017-02-25
Multi-touch

mars:
Hello creators of Popmotion ✨👆😍

I recently needed to discern multiple touches with Popmotion and so implemented a new `touches` action.

Demo CodePen: [Popmotion feature proposal: Touches](http://codepen.io/marsi/pen/Qdebvj/?editors=0010)

I'm currently using `touches` in production via scoped npm package: [@mars/popmotion](https://www.npmjs.com/package/@mars/popmotion). Try it:

```bash
npm install @mars/popmotion --save
```

I'd love to get your feedback about how this is implemented, and eventually see it included in Popmotion!


InventingWithMonster:
This is awesome man! This is one of those features that I wanted but never gotten around to implement so it's cool to see it actually real.

My only concern with the approach here is the difference between `changedTouches`, `targetTouches` and `touches` - to me this has always been an overly-complicated API. With `pointer` we reduce this down to `x` and `y`. Do you think there are practical reasons why we wouldn't reduce all the different touches down to a flat array of `x``y` points?

Playing around with it, it feels like by putting one finger down, then a second, and then moving both, `changedTouches` and `targetTouches` "overlap" on position `0`, whereas `touches` always works properly with multiple points. I feel like condensing this down into a flat array that Just Works, we could remove some of the usual confusion that comes with this event. (Or run the risk of breaking everything) - What do you think?

mars:
So happy to hear @InventingWithMonster 😄

I completely agree about condensing is down to a single list of coordinates. It will be better for Popmotion action API as well as performance (minimizing overhead in the event-to-action conversion).

The reason I enumerate all three TouchEvent coordinates today, is for observing, understanding & feedback about which coordinate is the right one to select.

`touches` does seem most accurate to me; always contains the touches I expect. The other coordinates `targetTouches` & `changedTouches` frequently omit additional touches on my iPhone.

I wonder if they behave differently on different platforms.

mars:
I updated to expose only the `touches`, avoiding the confusion of `targetTouches` & `changedTouches`.

Same Demo CodePen: [Popmotion feature proposal: Touches](http://codepen.io/marsi/pen/Qdebvj/?editors=0010)

InventingWithMonster:
Hey man sorry for the delay in response. 

This looks great although I'd go that last step and delegate `touchCount` to the consumer and simply output it as an array, as we have to do a lot less work to dig into the output and it's obvious what we should do if we want to know the length:

```
onUpdate: (v: Array<{x, y}>) => {
  const touchCount = v.length
}
```

With composabiliy in mind, it would remove the need for a mandatory 'pick' step, ie

```
onUpdate: flow(
  (v) => v.map(Math.round),
  (v) => console.log(v)
)
```

vs 

```
onUpdate: flow(
  ({ touches }) => touches,
  (v) => v.map(Math.round),
  (v) => console.log(v)
)
```

Also with `touches` now being an array we also make it easier to `map` like this.

What do you think about these proposals?

mars:
Thank for PR of all the revisions, @InventingWithMonster. Works great for me. Ready to merge.

InventingWithMonster:
@mars No worries, sorry it took this long!

The branch currently has conflicts so I can't merge it, could you take a look at these? I don't have the ability to directly contribute to your fork hence the round about way I did the last PR! They should be simple to resolve (the two `src` files will just need manual editing but the `dist` files can be replaced by fixing the `src` files and then running `npm run build && npm run watch`)

When I publish this, I'll add a `getChildren()` method to `parallel` and `composite` that returns an array/object respectively. That should clear up any funny terminology.

mars:
Okay @InventingWithMonster, I tried to take care of those `src/` file merge conflicts (which I did not inflict), but now the build is broken ending in the following error:

```
ERROR in Entry module not found: Error: Can't resolve './src/global-xl.js' in '/Users/mars/github/popmotion'
```

…looks like a recent devtools change on master that has left me unable to product the unminified bundle.

InventingWithMonster:
Ah my bad, that should be committed now.

mars:
Merged master again and successfully generated those dist files! Dare I say, all ready to go.

InventingWithMonster:
Awesome man! Thanks for your patience and work on this!

mars:
Thank you for creating & nurturing Popmotion @InventingWithMonster. So happy to see my contribution included.

PR #186, opened 2017-07-11
Allow calculating the velocity of a value over a window of times to g…

jacobp100:
In Leaflet, the inertial panning is done by averaging the values of `{ x, y }` and `time` over (in Leaflet's case) 50ms. This PR allows you to add this behaviour by calling `action.setVelocityWindow(50)`, but leaves the behaviour off by default.

We use two arrays rather than one array of `{ value, time }` objects to remove the need for garbage collection.

### DIFF
--- a/docs/api/action/README.md
+++ b/docs/api/action/README.md
@@ -21,6 +21,7 @@ category: action
 - `start`: Start an action. Fires any set `onStart` callbacks.
 - `stop`: Stop an action. Fires `onStop` callback.
 - `get`: Get the current value.
+- `setVelocityWindow`: Sets the duration of a window to average over for for velocity calculation, in milliseconds.
 - `getVelocity`: Get the action's current velocity, in units per second.
 - `getBeforeTransform`: Returns the current value, without passing through `transform` if a `transform` function is defined.
 - `complete`: Stop an action and mark as complete. Fires `onStop` and `onComplete` callbacks.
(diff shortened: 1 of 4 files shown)


InventingWithMonster:
Hey man

Thanks for the PR, I was only thinking about averaged velocities this morning. 

I'll give this a good read tomorrow, but I was wondering what advantages you find in this approach? I was thinking it'd be less suspectible to input noise?

jacobp100:
I think that's the idea—if you're tracking the mouse, you'll get a smoother velocity and a more accurate direction. To be honest, this approach is pretty much copy and paste from leaflet, I'm sure they'll have experimented with a a lot!

jacobp100:
A note on this is that we actually need to ensure that the values are added on each frame, which is not the case for a `value`.

InventingWithMonster:
Hey man sorry I haven't had the free time to look into this. I'm wondering if applying the smooth transformer on the velocity would provide similar results without retaining an array of previous velocities?

jacobp100:
It might do.

Did my previous comment make sense? I implemented this for inertial panning (like in Google Maps), but it had a bug. If you drag but hold the mouse for a second or so before releasing, the velocity is retained, so it starts inertial panning, even though the motion (to the user) had stopped.

Something that might work is to keep the value arrays, but when the value is updated, call `requestAnimationFrame` to update the velocity until the velocity reaches zero. This would probably have to be its own action because of the overhead (`trackVelocity`?)!

jacobp100:
I don't think this is the best way to fix this.

PR #227, opened 2017-12-14
Fix h4 skewing

pakastin:
When width is less than `1080px`, `<h4>` skew-correction doesn't work. This PR should fix that.
Even better would be to make code bg a separate div and only skew that.


InventingWithMonster:
Yeah agreed, apparently it's causing font rendering issues on Windows browsers. Thanks for this fix tho!

pakastin:
No worries, great lib – thank you! 🤘

PR #248, opened 2018-01-18
fix: unable to remove event listener on old chrome version

oliverzy:
This patch fix an issue related to listen action.
On old mobile chrome browser, listen.stop cannot remove the event listener when the options is set. It has the inconsistent behavior on the different browser.

Please see MDN doc related to `removeEventListener` for details.


InventingWithMonster:
Thanks man! I'll get this published as soon as possible

PR #261, opened 2018-03-04
Fix docs imports

brunnolou:
There is no default export to Calc, Easing and Transforms so the docs were wrong.


InventingWithMonster:
Nice one thanks for this!

PR #271, opened 2018-03-28
[Pose] Correct react children docs example + embed codepen example

jesstelford:
Feel free to fork the CodePen example and update the ID used here 👍


InventingWithMonster:
Thanks for your help man, I'll keep your version :)

PR #276, opened 2018-03-30
Support posing regular react components

jesstelford:
The same way Styled Components does it:

```javascript
class MyComponent extends React.Component {
  /* ... */
}

const DraggableComponent = posed(MyComponent)({
  draggable: 'x',
  dragBounds: { left: 0, right: 100 }
})

export default ({ radius }) => <DraggableComponent r={radius} />
```

I had some trouble getting the monorepo setup and running, but got there in the end - perhaps a setup guide might help future contributors?

Also, this is my second time ever writing TypeScript - I hope I didn't butcher it too bad!


packages/popmotion-pose/docs/api/react/posed.md
```diff
@@ -4,8 +4,6 @@ description: Create a posed component
 category: react
 ---
 
-**Note:** React Pose is built with React 16.3.0, which is currently in alpha.
```
jesstelford: 16.3.0 was released this morning 🎉

packages/react-pose/src/components/PoseElement.tsx
```diff
@@ -76,13 +79,18 @@ export class PoseElement extends React.PureComponent<PoseElementProps> {
     }
   }
 
+  getRef(): Element {
+    return this.hostRef || this.innerRef || this.ref;
```
jesstelford: Order here is important: Favouring `hostRef` (https://github.com/facebook/react/issues/11401) first, then any `innerRef` (for Styled Components support), then the original `ref`.

I'm not sure if the values could get out of date at all? Is there ever a case where the innerRef gets updated, but the hostRef doesn't? It all feels a bit icky, but less icky than using `.findDOMNode()` (which also doesn't work with Fragments).

packages/react-pose/src/components/PoseElement.tsx
```diff
@@ -199,17 +218,42 @@ export class PoseElement extends React.PureComponent<PoseElementProps> {
...
+            ...elementProps,
           },
           children
         )}
       </PoseParentContext.Provider>
     );
   }
 }
+
+reactLifecyclesCompat(PoseElement);
```
jesstelford: Give backwards compatibility with existing react code bases.

packages/popmotion-pose/docs/api/react/posed.md
arcticicestudio: I was about to create a PR for this right now :smile: 
But I think it should be better to only remove the part of the sentence behind the comma instead of removing the whole line. Otherwise some users might be confused when trying to use it with React < 16.3.

InventingWithMonster: @arcticicestudio This PR also adds the backwards compatibility polyfill so it should work with older versions of React

jesstelford:
Yeah, I went back and forward over using `findDOMNode`. I've given it another go just now, and ran into a couple of issues that need to be worked around / documented:

1. Can't use `findDOMNode` with functional components (I don't think there's any workaround for this beyond using a `innerRef` / `hostRef` prop)
2. Would have to document the need to pass `props.style` down to the rendered DOM node
3. Would have to document the (possibly) unexpected behaviour of `Fragment` components (`findDOMNode` will pick the first non-empty child).

A combination of points 2 & 3 could be very confusing - which element should receive the `props.style`? The work around there would be to avoid `Fragment` components all together (which is reasonable).

At this point, the documentation / API would become:

```md
#### Calling `pose` with a React component

`pose` is called directly, passing in [posed props](/pose/api/props):

``javascript
class MyComponent extends React.Component {
  /* ... */
}

const DraggableComponent = posed(MyComponent)({
  draggable: 'x',
  dragBounds: { left: 0, right: 100 }
})

export default ({ radius }) => <DraggableComponent r={radius} />
``

**Note:** React Pose requires a single DOM element to correctly calculate animations,
as such using `Fragment`s is not supported.

**Note 2:** Custom styled are applied to the DOM node being animated.
When calling `pose` with a React component, it is up to you to correctly apply `props.style` to the DOM node:

``javascript
class MyComponent extends React.Component {
  render() {
    <div ref={this.props.style}>Drag me</div>
  }
}
``

```

At this point, I feel we're in the same position as before, but using the non-recommended `findDOMNode` method as opposed to the RFC'd standard of `hostRef`.

Using `forwardRef` would clean up the `innerRef`/`hostRef` issue, but would require Styled Components to support it too. (and I can't find a polyfill, so it might be >=16.3.0 only)

I guess the question comes down to where you want the trade-offs: _"you must always pass `props.style`, and you can't use `<Fragment>`"_ vs _"you must always resolve `hostRef` (because Styled Components is weird with `innerRef`)"_?

jesstelford:
I've rebased onto `master` to resolve the merge conflict.

jesstelford:
I've published this version of the code to `@jesstelford/react-pose` if you want to try it out in a project.

InventingWithMonster:
@jesstelford thanks for that I'll give it a go now!

jesstelford:
👍 

I've rebased onto `master`, and published 57800027 as `@jesstelford/react-pose@1.1.1-anycomponent`

packages/react-pose/src/components/PoseElement.tsx
```diff
@@ -100,11 +108,11 @@ export class PoseElement extends React.PureComponent<PoseElementProps> {
     // If we're popping this element out from the DOM flow, build
     // and apply position: absolute styles that visually match the previous
     // location in the DOM
-    if (popFromFlow && this.ref && this.ref instanceof HTMLElement) {
+    if (popFromFlow && this.getRef() && this.getRef() instanceof HTMLElement) {
```
InventingWithMonster: Stick to `this.ref` - it doesn't matter where we got this from.

packages/react-pose/src/components/PoseElement.tsx
```diff
@@ -100,11 +108,11 @@ export class PoseElement extends React.PureComponent<PoseElementProps> {
...
       if (!this.popStyle) {
         props.style = {
           ...props.style,
-          ...calcPopFromFlowStyle(this.ref)
+          ...calcPopFromFlowStyle(this.getRef() as HTMLElement)
```
InventingWithMonster: As before

packages/react-pose/src/components/PoseElement.tsx
```diff
@@ -143,10 +159,10 @@ export class PoseElement extends React.PureComponent<PoseElementProps> {
 
     // If first in tree
     if (!registerChild) {
-      this.initPoser(poseFactory(this.ref, props));
+      this.initPoser(poseFactory(this.getRef(), props));
```
InventingWithMonster: Revert

packages/react-pose/src/components/PoseElement.tsx
```diff
@@ -117,14 +125,22 @@ export class PoseElement extends React.PureComponent<PoseElementProps> {
...
+
   setRef = (ref: Element) => {
     const { innerRef } = this.props;
     if (innerRef) innerRef(ref);
     this.ref = ref;
   };
 
   componentDidMount() {
-    if (!this.ref) return;
+    if (!this.getRef()) return;
```
InventingWithMonster: Revert

packages/react-pose/src/components/PoseElement.tsx
```diff
@@ -143,10 +159,10 @@ export class PoseElement extends React.PureComponent<PoseElementProps> {
 
     // If first in tree
     if (!registerChild) {
-      this.initPoser(poseFactory(this.ref, props));
+      this.initPoser(poseFactory(this.getRef(), props));
     } else {
       registerChild({
-        element: this.ref,
+        element: this.getRef(),
```
InventingWithMonster: Revert

packages/react-pose/src/components/PoseElement.tsx
```diff
@@ -168,6 +184,9 @@ export class PoseElement extends React.PureComponent<PoseElementProps> {
     const { onUnmount } = this.props;
     if (onUnmount) onUnmount(this.poser);
     this.poser.destroy();
+    this.ref = undefined;
+    this.innerRef = undefined;
```
InventingWithMonster: Revert

packages/react-pose/src/components/PoseElement.tsx
```diff
@@ -199,17 +218,42 @@ export class PoseElement extends React.PureComponent<PoseElementProps> {
...
+    // need to get a ref to an underlying DOM element. The standard is to accept
+    // an `innerRef` prop (ala: Styled Components), so we pass that function
+    // along.
+    // Unfortunately when using Styled Components, it will intercept the
+    // `innerRef` prop, and not pass it down to a wrapped component (when doing
+    // styled(Component)``, hence we pass in a _third_ ref getting function
+    // `hostRef`, which the consuming code can opt into if they want.
+    // ( see https://github.com/facebook/react/issues/11401 )
+    const elementProps = {
+      innerRef: this.setInnerRef,
```
InventingWithMonster:
```javascript
const elementProps = {
  innerRef: this.setRef,
  hostRef: this.setRef,
  ref: !isStatelessFunctionalComponent ? this.setRef : undefined
}
```

packages/react-pose/src/components/PoseElement.tsx
```diff
@@ -117,14 +125,22 @@ export class PoseElement extends React.PureComponent<PoseElementProps> {
     return props;
   }
 
+  setInnerRef = (ref: HTMLElement) => {
+    this.innerRef = ref;
```
InventingWithMonster: Remove both of these

packages/popmotion-pose/docs/api/react/posed.md
```diff
@@ -98,10 +135,10 @@ const sidebarProps = {
   closed: { x: '-100%' }
 }
 
-const Sidebar = styled(posed.nav(sidebarProps))`
+const Sidebar = posed(styled.nav`
```
InventingWithMonster: Leave this as before - I'd like to encourage the previous syntax because it leaves the complexity on Styled Components.

packages/popmotion-pose/docs/api/react/posed.md
```diff
@@ -26,6 +24,12 @@ import posed from 'react-pose'
 
 ### Create a posed component
 
+`pose` can be used in two ways:
```
InventingWithMonster:
```markdown
### Create a posed component

`posed` can be used to create posed components in two ways:
1) **Recommended:** Create HTML & SVG elements (eg `posed.div()`)
2) **Advanced:** Convert existing components (eg `posed(Component)()`)

#### HTML & SVG elements
````

File: packages/popmotion-pose/docs/api/react/posed.md [hunk 14]
```diff
@@ -37,6 +41,39 @@ const DraggableCircle = posed.circle({
 export default ({ radius }) => <DraggableCircle r={radius} />
 ```
 
+#### Calling `pose` with a React component
+
+`pose` is called directly, passing in [posed props](/pose/api/props):
```
InventingWithMonster: #### Existing components

Existing components can be converted to posed components by calling `posed` directly:

```javascript
const PosedComponent = posed(MyComponent)(poseProps)
```

For performance and layout calculations, React Pose requires a reference to the underlying DOM element. So, the component to be posed **must pass `hostRef` to the host element's `ref` prop**.

```javascript
const MyComponent = ({ hostRef }) => <div ref={hostRef} />

const PosedComponent = posed(MyComponent)({
  draggable: true
})

export default () => <PosedComponent pose={isOpen ? 'open' : 'closed'} />
```

For FLIP support in a `PoseGroup`, it **optionally** needs to pass on the `style` prop:

```javascript
const MyComponent = ({ hostRef, style }) => <div ref={hostRef} style={style} />
```

packages/react-pose/src/components/PoseElement.tsx
```diff
@@ -76,13 +79,18 @@ export class PoseElement extends React.PureComponent<PoseElementProps> {
...
+  }
+
   getSetProps() {
     const {
       children,
       elementType,
       poseProps,
       onChange,
       innerRef,
+      hostRef,
```
InventingWithMonster: Remove

packages/react-pose/src/components/PoseElement.tsx
jesstelford: Unfortunately, order matters with the ref. Check out [this example](https://codepen.io/jesstelford/pen/yKweKQ?&editors=0010):

```javascript
class SomeComponent extends React.Component {
  render() {
    return (
      <h1 ref={this.props.innerRef}>Hello World</h1>
    )
  }
}

class Posed extends React.Component {
  setInnerRef = () => {
    console.log('this.ref now points to <h1>');
  }
  setRef = () => {
    console.log('this.ref now points to <SomeComponent>');
  }
  render() {
    return (
      <SomeComponent ref={this.setRef} innerRef={this.setInnerRef} />
    )
  }
}

ReactDOM.render(
  <Posed />,
  document.getElementById('root')
);
```
Gives output:
```
this.ref now points to <h1>
this.ref now points to <SomeComponent>
```

Which brings us back to where we started (and the reason for this PR).

We could alias `this.hostRef` to `this.innerRef` without issue, so I'll make that change.

packages/popmotion-pose/docs/api/react/posed.md
jesstelford: Excellently worded, thanks!

jesstelford:
I've pushed those tweaks, but will have to work through the merge conflicts before this branch is ready again.

Thanks for the review!

The only outstanding issue now is: https://github.com/Popmotion/popmotion/pull/276#discussion_r180588433

packages/react-pose/src/components/PoseElement.tsx
InventingWithMonster: I’m not sure I fully understand the problem on this one? this.ref is an internal reference for us that we expose via hostRef. It’s up to the user to ensure that’s passed correctly the the host element rather than the component (if it’s using forwardRef for instance both of those refs will be h1)

InventingWithMonster:
Nice one thanks for taking another look at his!

packages/react-pose/src/components/PoseElement.tsx
InventingWithMonster: Oh I see. It’s late 😁
Maybe we should only pass on hostRef for now as per docs, maybe innerRef if we’re srue it won’t be a problem

InventingWithMonster: Turns out trying to get to sleep is great at figuring this shit out (again). I see, actually this time. Proceed. 🙃

InventingWithMonster: Wait wouldn’t it be possible to check if elementtype was a string and only send ref if it was, otherwise it’s a component and sent hostRef and innerRef?

jesstelford: Oh, of course! I'll give that a shot 👍

jesstelford: Your tired self has good ideas too ;)

jesstelford:
Quick update: Was banging my head on the desk for many an hour trying to figure out why I was getting a `unexpected fiber popped` error. Turns out it was a React 16.3.0 bug. Updating my test codebase to 16.3.1 fixed it.

Am attempting the rebase / checking if the type is a string now.

jesstelford:
Phew! What a journey. Ok, got the changes pushed.

Have published it as `@jesstelford/react-pose@1.4.0-anycomponent` if you want to try it out 👍

InventingWithMonster:
You’re an actual legend. I installed Leona but never took advantage. And broke all our code pens with that React bug, that was a sweaty moment 😅

I’ll merge and release at some point today, thanks for your hard work on this

jesstelford:
Haha, thanks mate :)

Be sure to test it out with any example apps, etc, you've got - I don't want to have crept any bugs in there (especially since I didn't write any tests).

jesstelford:
🎉

jesstelford:
@InventingWithMonster just gave `react-post@1.5.1` a go, but am getting errors with the code setup I was using. Some investigation shows that a bunch of the code got clobbered in 69657214e722f7aacf89db5d684be8e50c9653c0 - not sure if this was intentional?

Here's a quick example showing the errors: https://github.com/jesstelford/react-pose-issue

It also throws warnings when posing a functional component (since it tries to attach the `ref` prop).

InventingWithMonster:
Hey man thanks for the working demo, il take a look at work. I was playing
around seeing how little code I could get away with for this, and that
seemed to work pretty well. Also the innerRef stuff I decided was always
going to be used to fish out the underlying DOM node (as per my head docs
as I forgot to add them to the actual docs) which is why that logic went

What’s happening here is (as I’m sure you encountered) one of the refs is
coming through as a React component. I’ll look into why. I also don’t
understand the functional component error as now the only time we sent ref
is if elementType is a string. I’ll put back the explicit functional
component check

InventingWithMonster:
@jesstelford Hey man

Fixed the error with https://github.com/Popmotion/popmotion/commit/9a1027e29ae18a6b6ff0ef14824e2d6ddd4ca9d9#diff-57a535dfcce9662e7d53f9cdff1b0073R163

As a heads up the functional ref warning is still there, which I suspected as I don't think there's a way of detecting a Styled Component functional component and was still getting with `@jesstelford/react-pose`

PR #277, opened 2018-03-30
Conditional rendering inside PoseGroup

efimweb:
This pull request enables to use conditional rendering in PoseGroup. For example, for [button icon which depends on state](https://codepen.io/efimweb/pen/yKKXYJ?editors=0010)


jesstelford:
:+1:

Needs to also be added to the `animateChildrenList`'s `Children.forEach`, and `mergeChildren`'s `incomingChildren.forEach` in the same file.

efimweb:
What for? These functions process children, that have already been filtered in `PoseGroup.getDerivedStateFromProps` with `utils.makeChildList`

jesstelford:
D'oh, you're right, sorry. I copy+pasted the diff here into my local copy, and ran it without rebuilding, and thought spots were missed!

PR #279, opened 2018-04-01
Fix #267 - staggerDirection: -1

### LINKED ISSUE #267: Pose: `staggerChildren` not working when `staggerDirection` is -1
Seems like this line https://github.com/Popmotion/popmotion/blob/master/packages/popmotion-pose/src/factories/pose-setter.ts#L42

Isn't correctly inverting the stagger times

efimweb:


InventingWithMonster:
Excellent man thanks!

PR #296, opened 2018-04-12
popmotion-spinnable - add setRotate method

gpmcadam:
adds a `setRotate` method to popmotion-spinnable, which allows overriding the result after initialisation


InventingWithMonster:
Closed for inactivity

PR #309, opened 2018-04-26
Add Ol component to the documentation

EmilTholin:
There is currently no component for `ol` elements, so the FLIP page looks like this:

flip_current

This PR adds a `Ol` component that is used for `ol` elements, identical to the `Ul` component except that the `list-style-type` is decimal:

flip_new


InventingWithMonster:
Legend thanks

Issue #274, opened 2018-03-29
`steps` returns NaN if range is 0,0


InventingWithMonster:

InventingWithMonster:
https://github.com/Popmotion/popmotion/pull/311

PR #311, opened 2018-04-26
Handle step ranges of length 0

EmilTholin:
Step ranges of length 0 results in `getProgressFromValue` dividing by 0, which makes the function returned from `steps` return `NaN` for every step given to it.

Returning a progress of `1` when the range is of length 0 resolves the issue. 

Closes https://github.com/Popmotion/popmotion/issues/274


InventingWithMonster:
This is awesome man thanks

PR #340, opened 2018-05-20
Tweaked build setups - omit rollup warnings, annotate pure calls & more

Andarist:
Please review this carefully, I've done those changes quickly & I have no actual knowledge about this project (I have not yet used it, although I'm planning to 😉 ). As always - if something is unclear, please ask about it and I'll try my best to explain the change in question.


InventingWithMonster:
I've merged but gonna have to remove the `postbuild` step  - as far as I can tell it just annotates the pure functions but uglify isn't run so I don't *think* this affects the output minified file?

InventingWithMonster:
Oh I see this is more for when people consume it as a package? I'll leave it as is for now but will look into a rollup step further down the line to consolidate this logic

InventingWithMonster:
PS Thanks for the PR 😅

Andarist:
> Oh I see this is more for when people consume it as a package?

Yeah, exactly 👍

PR #349, opened 2018-05-23
babel-plugin-popmotion

Andarist:
Idea is that using `posed.div` provides a nice DX, but it costs both initialization time (calling `.reduce`) & keeping `supportedElements` whitelist array in the bundle. 

We can replace all `styled.div({})` to `styled('div')({})` at build time and rewrite imports to `react-pose/base` (new entry) to remove this `supportedElements` from final applications.

### DIFF
--- a/packages/babel-plugin-popmotion/jest.config.js
+++ b/packages/babel-plugin-popmotion/jest.config.js
@@ -1,0 +1,5 @@
+module.exports = {
+  transform: {
+    '\.js$': '<rootDir>/scripts/babel-jest-transform.js'
+  }
+}
--- a/packages/react-pose/es/base.js
+++ b/packages/react-pose/es/base.js
@@ -1,0 +1,374 @@
+import { __rest, __assign, __extends } from 'tslib';
+import { createElement, createContext, Fragment, Component, PureComponent, Children, cloneElement } from 'react';
+import poseFactory from 'popmotion-pose';
+import { warning, invariant } from 'hey-listen';
+
+var PoseParentContext = /*#__PURE__*/createContext({});
+var calcPopFromFlowStyle = function (el) {
+    var offsetTop = el.offsetTop,
+        offsetLeft = el.offsetLeft,
+        offsetWidth = el.offsetWidth,
+        offsetHeight = el.offsetHeight;
+    return {
+        position: 'absolute',
+        top: offsetTop,
+        left: offsetLeft,
+        width: offsetWidth,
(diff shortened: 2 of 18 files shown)


packages/babel-plugin-popmotion/jest.config.js
```diff
@@ -0,0 +1,5 @@
+module.exports = {
+  transform: {
```
Andarist: because this plugin tests against both babel@6 and babel@7 I was more comfortable with providing the babel transform for jest explicitly instead of relaying on its automatic behaviour

packages/react-pose/es/base.js
```diff
@@ -0,0 +1,374 @@
+import { __rest, __assign, __extends } from 'tslib';
```
Andarist: personally I would `.gitignore` this file, but similar directories (`dist`) in other packages in this repo are not ignored, so I left it for now.

packages/react-pose/rollup.config.js
```diff
@@ -59,21 +59,26 @@ const umdProd = Object.assign({}, umd, {
...
+  experimentalCodeSplitting: true,
   output: {
-    file: 'lib/index.js',
+    dir: 'lib',
     format: 'cjs',
-    exports: 'named'
+    exports: 'named',
+    entryNames: '[alias].js'
   },
+  // plugins: [typescript(declarationsConfig)]
```
Andarist: unfortunately both this config and umd builds are failing due to some typing issues, could u check them out? I'm not well versed in TS and I have no idea why they are failing

InventingWithMonster:
Hey man thanks for the PR but I'm going to decline this one for the specific reason that I won't have the capacity to support it. I think the idea is great though, I feel like it could be expanded to support both `react-pose` and `styled-components`?

Andarist:
> Hey man thanks for the PR but I'm going to decline this one for the specific reason that I won't have the capacity to support it.

Is this solely because of lack of experience in AST transformations? I can commit to maintaining this, especially that this transform is minimal and doing really easy transformations.

>  I think the idea is great though, I feel like it could be expanded to support both react-pose and styled-components and released as a stand-alone project?

`styled-components` already has its own plugin. Those transforms done here could be indeed abstracted and reused in both projects, but even with that each library should provide its own plugin which would basically just call abstracted "visitor" (transforms) factory with its own configuration (like import names etc). So some maintenance (minimal) would still be on each library's maintainer.

PR #356, opened 2018-06-01
only call onDragEnd once

spruce-bruce:
I was playing with react native draggable components and I noticed that `onDragEnd` is being called twice. I took a look at the source and saw that it was explicitly being called twice, and I think that's unintentional. Here's a PR to get rid of that extra call

### DIFF
--- a/packages/react-native-pose/src/index.ts
+++ b/packages/react-native-pose/src/index.ts
@@ -97,7 +97,6 @@ const posed = createPosed({
         if (onDragEnd) onDragEnd(e, gestureState);
         if (dragX) values.x.flattenOffset();
         if (dragY) values.y.flattenOffset();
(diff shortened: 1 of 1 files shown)


InventingWithMonster:
Legend thanks, yeah that is indeed what we call in the business, a fuck up

InventingWithMonster:
I'm away today but will try and get this published tomorrow

spruce-bruce:
I'm not stranger to fuck ups - having a good time with the lib btw, thanks for the work!

beaurushton:
Any chance we can get this published soon?  Just went through the process of hunting down this bug myself only to find my way here to see that it has already been fixed and the PR merged.  Be great if we can get an updated package.  Thanks for the fantastic library!

PR #364, opened 2018-06-07
Remove unused dev dependencies

voxpelli:
Looks like `ts-loader` stopped being used when Rollup was introduced and `ts-list` I can't find within the code base either


InventingWithMonster:
Nice one thanks

PR #389, opened 2018-07-02
Add document touch listeners lazily

voxpelli:
Only add document touch listeners when anyone is actually interested in them – it feels pretty wasteful to add such listeners whenever the Popmotion library is included, no matter if one uses the code or not (could be solved by doing some tree shaking, but not everyone has a build environment set up for that, eg. is using commonjs and Browserify)

This should have no real side-effects as `updatePointsLocation()` replaces the content of  `points` every time it's invoked – so there's really no need to track anything prior to the creation of a multitouch input and since its still ensured that only one global listener is added.

Only small edge case is if a touch is already happening when the multitouch input is happening – then that touch won't be detected until the next `touchstart` or `touchmove` – but I think that is acceptable considering the benefits of not _always_ having touch listeners when one includes Popmotion, no matter if one uses the multitouch input or not?

### DIFF
--- a/packages/popmotion/src/input/multitouch/index.ts
+++ b/packages/popmotion/src/input/multitouch/index.ts
@@ -41,6 +41,9 @@ const removePointsListener = () => {
     if (pointsListener && pointsListener.stop) {
       pointsListener.stop();
       pointsListener = undefined;
+
+      points.splice(1);
+      points[0] = defaultPointerPos();
     }
   }
 };
--- a/packages/popmotion/src/input/multitouch/index.ts
+++ b/packages/popmotion/src/input/multitouch/index.ts
@@ -4,27 +4,46 @@ import { angle, distance } from '../../calc';
 import listen from '../listen';
 import { PointerPoint, PointerProps } from '../pointer/types';
 import { defaultPointerPos, eventToPoint } from '../pointer/utils';
+import { ColdSubscription } from '../../action/types';
 
 const points: PointerPoint[] = [defaultPointerPos()];
 let isTouchDevice = false;
 
-if (typeof document !== 'undefined') {
-  const updatePointsLocation = ({ touches }: TouchEvent) => {
-    isTouchDevice = true;
-    const numTouches = touches.length;
+let listenerCount = 0;
+let pointsListener: ColdSubscription | void;
 
-    // TODO: Optimisation would be to provide existing points to `eventToPoint`
-    points.length = 0;
+const updatePointsLocation = ({ touches }: TouchEvent) => {
+  isTouchDevice = true;
+  const numTouches = touches.length;
 
-    for (let i = 0; i < numTouches; i++) {
-      const thisTouch = touches[i];
-      points.push(eventToPoint(thisTouch));
+  // TODO: Optimisation would be to provide existing points to `eventToPoint`
+  points.length = 0;
+
+  for (let i = 0; i < numTouches; i++) {
+    const thisTouch = touches[i];
+    points.push(eventToPoint(thisTouch));
+  }
+};
+
+const addPointsListener = () => {
+  listenerCount += 1;
+
+  if (typeof document !== 'undefined' && !pointsListener) {
+    pointsListener = listen(document, 'touchstart touchmove', true)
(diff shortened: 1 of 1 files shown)


voxpelli:
@InventingWithMonster Anything I can do to aid in the reviewing of this PR? Looking forward to get this contribution in + to follow up with more as we continue to use Popmotion at @sydsvenskan

InventingWithMonster:
Hey @voxpelli 

Thanks for the contributions and sorry for the delay on this one, slammed for time recently.

Sadly though I've decided not to merge this one.

Originally, we had code like this because I also had the gut feeling that it was "wrong" or wasteful as you say, to add an event listener, even if the user isn't using `pointer`.

But I ended up writing it like this primarily because there's no detectable overhead even on low power devices for adding a single event listener. So to keep a running count of listeners and actively manage DOM listeners ends up actually just complicating the code, leaving it more brittle.

There's even the possibility that adding and removing listeners has a higher runtime cost over the duration of the run time and incurred GC could negatively affect animations (but again, I imagine this would be undetectable even on low power devices and probably isn't argument for or against)

In the end I decided that I'd written a micro-optimisation that'd only have true benefit if this block of code was repeated hundreds or thousands of times and deferred towards code clarity.

Again thanks for the contribution, it is appreciated even if I'm not merging this one in.

voxpelli:
@InventingWithMonster While an event listener in itself won't have any real impact on anything, there are a few events that when listened to non-passively disables a couple of optimizations within the browser and my impression is that touch listeners are such events.

See eg. this quote from https://developers.google.com/web/updates/2016/06/passive-event-listeners:

> Many times when you encounter scroll jank, the culprit is a touch event listener. [...] Currently, browsers can't know if a touch event listener is going to cancel the scroll, so they always wait for the listener to finish before scrolling the page.

I would prefer the listener count, but at the very least I would like to see the listener initiated lazily when the first touch listener is added – and not be added whenever someone includes the popmotion library for non-touch purposes. Especially so if the performance impact is true.

The performance impact right now can't be measured in eg. Chrome as Chrome decided to deal with this by making such listeners passive by default: https://developers.google.com/web/updates/2017/01/scrolling-intervention

One can see on MDN which browsers that has decided to go against the specification and default `touchstart` and `touchmove` to passive mode: https://developer.mozilla.org/en-US/docs/Web/Events/touchstart (See also: https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener#Improving_scrolling_performance_with_passive_listeners) Eg. Edge, IE and Safari still stick to specification and Firefox made them passive first in Firefox 61, which is the current stable release of it.

InventingWithMonster:
That permanent touch listener _is_ passive though, so it won't prevent these browser optimisations in supported browsers.

The other thing is that by keeping a permanent pointer state it allows us to cut a frame of lag when calculating deltas. Back when we were actively tracking pointers we required the user to pass through the originating `event` object (usually `mousedown` or `touchstart`) (`pointer(e)`)

This allowed us to use that event as the start point, so calculating relative offsets could begin immediately. 

With Popmotion 8 I went hard on simplifying the API and that included removing this requirement. Now, without the starting event we don't have access to that initial pointer position (unless we constantly track it). So we have to wait an additional event before figuring out offsets. Indeed this PR would reintroduce that initial frame of lag because we wouldn't be updating the pointer position until after the `touchstart` event had been disposed of.

voxpelli:
@InventingWithMonster This PR aside, are you absolutely sure that the touch listener is passive? I'm not seeing that. See eg. this annotated screenshot of the Chrome debugger:

group

`target.addEventListener(type, listener[, options]);` is being sent `true` as `options`, which is the legacy way of sending `{ capture: true }` and what needs to be sent to make a passive listener is `{ passive: true }`?

Firefox 61 and later + Chrome 55 and later will as I said earlier make these events passive anyway, but that's to bypass the often missed `{ passive: true }` option that both Edge and Safari still requires and which the spec still requires.

So either you or I am missing something here? I hope it's me 😝

InventingWithMonster:
Hm good catch yeah it won't be in Edge and Safari - https://github.com/Popmotion/popmotion/issues/412

PR #399, opened 2018-07-17
Fix hasChanged returning false when array items change

dannsam:
Just found this issue while playing with react-pose.
When change some items in the array of poses, if the number of items is the same, the new pose does not work, eg:
```tsx
<PosedDiv pose={["flip", expanded ? "expanded" : "collapsed"} />
```


InventingWithMonster:
Excellent thanks man!

PR #411, opened 2018-08-04
Extracted @popmotion/easing package

Andarist:
This is probably not yet ready to merge - I'm not sure how to properly test this in this monorepo setup.

**Why?**
Those easing functions are extremely useful & I'd like to import a single one and let tree-shaking to its thing to leave only the imported one in my bundles, but it's not possible in current setup - because I have to import whole `easing` object from `popmotion` library (which holds also higher risk that some other things from that lib might not get tree-shaken properly and would leave me with bigger bundles).

### DIFF
--- a/packages/easing/package.json
+++ b/packages/easing/package.json
@@ -1,16 +1,16 @@
 {
-  "name": "popmotion",
-  "version": "8.3.3",
-  "description": "A functional, reactive motion library.",
+  "name": "@popmotion/easing",
+  "version": "0.0.0",
+  "description": "A set of modular easing equations.",
   "author": "Matt Perry",
   "homepage": "https://popmotion.io",
   "main": "./lib/index.js",
   "types": "./lib/index.d.ts",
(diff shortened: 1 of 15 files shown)


InventingWithMonster:
Hey man! I like this a lot, good call. I'll try to merge and publish this soon.

Andarist:
Please recheck my work, it was kinda YOLO without proper testing 😉 Thanks for merging! In general this could be done for other things - calc? transformers? I just did this only for easing because it was the one that I've wanted to reuse

InventingWithMonster:
Yeah I think the others are a great candidate for similar treatment. I've published this now - though I won't be able to update Popmotion itself until next Weds.

Andarist:
No problem, I'm not in a hurry 😉 For now I've "unmangled" a single easing to:
```js
const easeOut = percent => 1 - (1 - percent) * (1 - percent)
``` 
and works great !

PR #413, opened 2018-08-07
Rename noop to identity

Andarist:
Imho `noop` is more often used for `() => {}`, while `identity` is a common name for `a => a`


InventingWithMonster:
Looked up the definition, liked it!

PR #414, opened 2018-08-07
Tweaked appendUnit JSDoc

Andarist:
I'm not a JSDocs user, so this might be slightly wrong - I've used their documentation and some github issues to fix this.


InventingWithMonster:
Nice one thanks!

PR #416, opened 2018-08-11
react-pose: Fix component argument type in `posed`

strax:
The `component` argument in `posed` was `React.Component` which would only allow _instances_ of `React.Component` subclasses. [The documentation][1] states that a component constructor needs to be passed to `posed` for custom components, so the type needs to be `React.ComponentType` to allow class component constructors and stateless components.

[1]: https://popmotion.io/pose/api/posed/#{{0}}-usage-create-a-posed-component-existing-components

### DIFF
--- a/packages/react-pose/src/posed.tsx
+++ b/packages/react-pose/src/posed.tsx
@@ -13,7 +13,7 @@ export type ComponentFactory = (
 ) => (props: PoseElementProps) => ReactElement<any>;
 
 export type Posed = {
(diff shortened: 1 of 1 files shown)


InventingWithMonster:
Excellent thanks for this!

InventingWithMonster:
Hey @strax, I've had to revert this commit as it was breaking our builds. I need to add CI Circle or something but in the meantime you can check this by running `yarn` in root and then `yarn build` in the specific directory

PR #417, opened 2018-08-11
[Documentation] fix link targets for links to stylefire docs

hputzek:
Links to stylers in the stylefire documentation point to the right URL now.

You can test the current (wrong) state on the live site:

* go for example to: https://popmotion.io/api/svg/
* try clicking on "Moved to stylefire docs"


InventingWithMonster:
Nice one thanks for this!

PR #424, opened 2018-08-14
correctly check the current `poseKey` against the previous `poseKey`

klaasman:
See the diff, it's pretty self-explanatory.


InventingWithMonster:
Nice one thanks for this!

InventingWithMonster:
Published @2.2.2

PR #427, opened 2018-08-16
Initial lerna setup

### LINKED ISSUE #426: Lerna
Would u be interested in me setting this up for you? Would probably make working with this repository easier - as lerna has some nice features like i.e. running a script in every package at once.

Andarist:
Closes #426 

This is obviously just a simple setup, without much in it. All the rest depends on your needs - tell me about your monorepo pain points and I'll try to do my best to ease them.

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -1,0 +1,7 @@
+{
+  "lerna": "3.0.6",
+  "version": "independent",
+  "packages": ["packages/*"],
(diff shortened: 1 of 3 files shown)


InventingWithMonster:
This is awesome thanks. The main thing I think is a problem is running yarn build in every directory when you first clone it, a command to do that would be great. I haven't encountered more than that I don't think

PR #428, opened 2018-08-17
Rewrite testVectorProps to use []#some instead of []#reduce

Andarist:
This is both shorter & should be able to exit the iteration sooner.


InventingWithMonster:
Ah nice catch 👍

Issue #304, opened 2018-04-20
Props being passed down to DOM element


juhaelee:
It appears that invalid props are being passed to the DOM when creating React components used posed. For example:

> Warning: React does not recognize the `dragBounds` prop on a DOM element. If you intentionally want it to appear in the DOM as a custom attribute, spell it as lowercase `dragbounds` instead. If you accidentally passed it from a parent component, remove it from the DOM element.

sonaye:
Yes this is an annoying issue, makes debugging difficult. Here is a workaround I've been using:

```javascript
// when you have
<Foo bar="baz" />
// created with
const PosedFoo = posed.div({ /* .. */ });
const Foo = styled(PosedFoo)`/* .. */`;
// it will scream a warning at you (React does not recognize the `bar` prop ..)

// to block the `bar` prop completely
const Foo = styled(({ bar, ...props }) => <PosedFoo {...props} />)`/* .. */`;
// however this may end up breaking your styles or animations

// solution: use a data attribute as a prop `pose-something`
<Foo pose-bar="baz" /> // gotcha: value must be a string
// instead of `props => props.bar` use props => props['pose-bar']
// keep in mind it will still show `pose-bar` in the dom
```

InventingWithMonster:
For my (or anyone who fancies writing a PR's) future reference

https://github.com/styled-components/styled-components/blob/master/src/utils/validAttr.js

Leeds-eBooks:
This also happens in `react-native-pose`:

```js
const Posed = posed.View({
  loading: {
    x: ({orientation}) =>
      getCenterPosition(
        orientation === 'portrait' ? screenWidth : screenHeight
      )
  },
  loaded: {x: 0},
})

<Posed
  orientation={props.orientation}
  pose={props.loaded ? 'loaded' : 'loading'}
  style={styles.posed} />
```

The output becomes:

```js
<View
  collapsable={undefined}
  orientation="portrait"
  style={
    Object {
      ...
    }
  }
/>
```

Currently, passing the prop `orientation` down to the actual `View` doesn't cause any problems, but that's purely a happy accident, as someone could easily use a prop name that does have an effect on the underlying `View`.

InventingWithMonster:
Exactly yeah - this is the double edged sword. I'm treading lightly here.

There is a `posed(Component)` API but it's not quite as simple to use. The one design goal of Pose is simplicity.

[...]

PR #429, opened 2018-08-17
Delete invalid DOM attributes (#304)

klaasman:
I'm not sure whether you should merge this pull-request or not: it does fix the issue but I feel more for the `transitionProps`-solution as given by @Leeds-eBooks (https://github.com/Popmotion/popmotion/issues/304#issuecomment-397240840).


Leeds-eBooks:
Would this work with React Native?

klaasman:
Good point. The attribute validation function is borrowed from styled-components, which is react-native compatible. We'd need to dive into their codebase to see if there are other checks built in for native.

InventingWithMonster:
@klaasman great PR! Looking forward to this. In SC do they outright delete invalid props or do they somehow not add them in the first place?

Leeds-eBooks:
The more I think about it, the more I think there should be a `transitionProps` prop. Maintaining a list of valid DOM attributes is ok because they change rarely and everyone knows about upcoming changes to HTML. RN is a different beast, and having to maintain a list of accepted props for the various native components will be hard with such a fast-moving platform. Plus, asking devs to be explicit about which props they want passing down and which they only want to be used for pose functions is no bad thing in my opinion. If you want a prop to be used in both a pose function and to be passed through, you can just put it twice.

InventingWithMonster:
@klaasman this is the approach I was thinking of https://github.com/styled-components/styled-components/blob/7aaf843b3983c7af9583c71a773a26341bfaf748/src/models/StyledComponent.js#L247

InventingWithMonster:
I'd assume this works with RN but we'll have to test it out. RN stuff lives in `react-pose-core` anyway

klaasman:
> The whole "picking" process should be also more efficient than copying first + deleting afterwards

Isn't your solution effectively doing the same thing, but then with a comparator for every key? (which sounds less efficient)

Nevertheless, your solution is fine too. With the default `isPropValid`-check it has the same behavior with the aforementioned additional benefits.

klaasman:
There's this prop called the `poseKey`, which can be changed to re-trigger functions in the pose configuration. I'm thinking it might be possible to ditch this prop if we go for the `transitionProps` solution. Any thoughts on this?

Andarist:
You iterate through all props when doing the initial merge (`{ ...this.getSetProps(), ...this.getRefs() }`), later you iterate through them again with `forEach` and you have to test each prop just like me to delete invalid ones. In general from what I know it's better to avoid deleting properties because it messes up hidden classes in JS engines (this might be outdated slightly at this point, but following a little what's going on in Chrome this should still be true - at least to some extent, maybe they are mitigating this deopt somehow nowadays).

My solution is ofc really similar - I just do a one pass iteration with `pickAssign` and the comparator test is done within this iteration. Additionally because `isValidProp` is a memoized function the comparator tests should be faster with time (because they will be looked up in a hash object instead of retesting the inputs each time with regexp).

As to `transitionProps` - personally I don't quite like it. IMHO makes the API more clunky *and* is easier to deopt the posed component (which extends PureComponent) by passing a new object reference in each render of the consuming component.

InventingWithMonster:
Is isValidProp being memorised a potential source of server-side memory
leaks?

I agree that transitionProps is not my preferred solution. It feels
arbitrary and it’s unlikely we’ll be able to do a shallow compare as it
lends itself to setting as a new object in render. Which is why I think
it’s unlikely we’d be able to use it as a replacement for poseKey

Andarist:
> Is isValidProp being memorised a potential source of server-side memory

The memoization is not weak, so theoretically this might be slight concern - but it shouldnt grow too much if you do not use thousands of different props (unlikely) and still the memory cost would be negligible. It's used in emotion which is popular css-in-js lib and we had no complaints about this and I know people are doing SSR with emotion.

Leeds-eBooks:
>it’s unlikely we’ll be able to do a shallow compare as it lends itself to setting as a new object in render.

Given that `transitionProps` would be a direct replacement for props passed directly into the element, why couldn't you just shallow compare the top-level properties of the `transitionProps` object? You might thereby be able to deprecate the `poseKey` prop?

klaasman:
I have to admit that transitionProps isn't my preferred solution either, bus it has its benefits. I think the shallow compare is possible, as @Leeds-eBooks states.

Another advantage is the freedom of how to name your props, e.g. currently the prop `value` will never work since it will be overridden by pose with a value object.

Leeds-eBooks:
>Would break PureComponent's shallow compare

Is your `Box` element a posed component (e.g. a `posed.div()`)? If so, isn't it simply up to Pose to handle the `shouldComponentUpdate` logic? I think I must be missing something obvious!

Andarist:
> Is your Box element a posed component (e.g. a posed.div())?

yeah

> If so, isn't it simply up to Pose to handle the shouldComponentUpdate logic? I think I must be missing something obvious!

Custom `shouldComponentUpdate` is an option. would have to change the base class of PoseElement to React.Component (from React.PureComponent), but that's minor thing. Seems to me like kinda reinventing the wheel when we can have shallow compare "for free" when using `PureComponent`. This has other costs though - keeping the props whitelist, filtering etc so maybe it would be worth it.

InventingWithMonster:
@klaasman Thanks for this PR, I've been really pleased that the community has come together to tackle this one. But I'm going to close it in favour of [#431](https://github.com/Popmotion/popmotion/pull/431).

Reason being, from an API point of view these PRs are equivalent. But the other PR is using `@emotion/is-prop-valid` which will reduce filesize for users of both Pose and Emotion. It's also more maintainable than a regex. It might be possible for us to open a PR for Styled Components to see if they'd use the module too, so we can reduce filesize for more users.

To answer the other subject in this discussion - `transitionProps` is more explicit and would neatly sidestep this entire problem, but it brings complexities to the API which is against the single core principle of Pose (simplicity).

It also doubles up the props comparison, reimplementing something that React already does for us.  

It's true that the approach doesn't fix React Native, but then the userbase for React Native Pose is currently very low and it also uses its own version of `react-pose` anyway. It'd be possible, if no other solution was found, to implement `transitionProps` there. But I feel like it won't be necessary.

Andarist:
> It's also more maintainable than a regex. It might be possible for us to open a PR for Styled Components to see if they'd use the module too, so we can reduce filesize for more users.

I'm planning to do it, a standard way of filtering the non-valid props would be beneficial to other libraries as well. Slight problem with `styled-components` is that their regexp is a little bit more complicated (i.e. validating the first character of data attributes) - IMHO it gives no real benefit, but ofc increases correctness, but sadly switching to other filtering function would a breaking change (sort of) for them. I know v4 is coming so maybe we could use that timing to start a discussion around this.

> It's true that the approach doesn't fix React Native, but then the userbase for React Native Pose is currently very low and it also uses its own version of react-pose anyway. It'd be possible, if no other solution was found, to implement transitionProps there. But I feel like it won't be necessary.

Emotion has similar testing function [here](https://github.com/emotion-js/emotion/blob/8edf5d33fe5ed26d3cc1bab24c03b0445e8cff56/packages/primitives/src/test-props.js), we could package it up too and unify React DOM & React Native pose libraries. The only difference would be which testing function gets configured as default for each one.

klaasman:
@InventingWithMonster No problem! I'm glad there's movement with this issue 👍

Andarist:
Just un update - styled-components are switching to `@emotion/is-prop-valid` too! https://github.com/styled-components/styled-components/pull/1937/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R65

PR #432, opened 2018-08-19
Keep codesandboxes source in repo

Andarist:
This should allow for:
- easier development
- keeping them up to date (auto update from repo to codesandbox)
- easier example PRs

### DIFF
--- a/examples/pose-image-filter/package.json
+++ b/examples/pose-image-filter/package.json
@@ -1,0 +1,22 @@
+{
+  "name": "react-pose-image-filter",
+  "description": "",
+  "private": true,
+  "main": "src/index.js",
+  "homepage": "https://codesandbox.io/s/github/Popmotion/popmotion/tree/master/examples/pose-image-filter",
+  "keywords": [],
+  "dependencies": {
+    "react": "16.3.2",
+    "react-dom": "16.3.2",
+    "react-pose": "latest"
+  },
+  "devDependencies": {
+    "react-scripts": "^1.1.4"
+  },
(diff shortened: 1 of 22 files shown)


examples/pose-image-filter/package.json
```diff
@@ -0,0 +1,22 @@
+{
```
Andarist: ⚠️ this one is not working at the moment, but it's because the original sandbox got tweaked with new `hoverable` changes but it's not yet working because of `default` keyword usage

InventingWithMonster:
Hey man thanks for this PR, sorry I didn’t see it before going a bit wild on the examples today! My concern with this approach is it prevents just forking/spinning up new sandboxes in CodeSandbox and using their editor? Or can those get commited to our repo?

Andarist:
> Hey man thanks for this PR, sorry I didn’t see it before going a bit wild on the examples today!

No worries, if you like I will just rebase & tweak newly added examples.

> My concern with this approach is it prevents just forking/spinning up new sandboxes in CodeSandbox and using their editor? Or can those get commited to our repo?

It will still be possible to just fork a sandbox. It's even possible to create PRs to upstream repo from such codesandboxes, I'm not sure how this works as I haven't used this feature but you can read more about it [here](https://codesandbox.io/docs/git)

InventingWithMonster:
Hey man thanks for the PR but yeah I think for now I'm not going to keep the examples in the repo, it might happen in the future but I think I've got a good workflow going at the moment with minimal overhead.

PR #434, opened 2018-08-20
Fix react-pose ComponentFactory parameter type

dannsam:
`React.Component` represents an instance of component class.
`React.ComponentType` needs to be used instead to support
stateless components and class components.
This change fixes typing issues in examples like:
```
const MyComponent = ({ hostRef }) => <div ref={hostRef} />

const PosedComponent = posed(MyComponent)({
  draggable: true
})
```
Without this fix, there's an error `Property 'setState' is missing in type...`


InventingWithMonster:
That's published as `react-pose@3.0.1`

PR #443, opened 2018-08-26
Monorepo alias

Andarist:
By aliasing lerna packages u can now build:
- all packages at once from the root directory
- each package separately - and they will always use "latest" code of other packages in the monorepo

The same thing can be later done for test scripts.


InventingWithMonster:
Hey man thanks for this, just one question what's the thinking behind not publishing declaration files and just linking through to the typescript source? Does it make any difference?

Andarist:
It works, but supposedly not the best idea - https://twitter.com/drosenwasser/status/1034357315057549313

Need to restructure this part of the changes, so it's not quite ready to merge yet 😢

InventingWithMonster:
Hey @Andarist do you mind if I close this PR?

Issue #265, opened 2018-03-20
Mouse doesnt work after touch is used in Pointer.


gshireesh:
all the demos in popmotion.io fall under this.

example:
https://codepen.io/popmotion/pen/RjBZoe?editors=0010

my environment:
Ubuntu, on chrome. dell xps 13(touch enabled)

**how to reproduce issue:**
move the ball with touch enabled screens, and then try to move with mouse. 

**this case works**
move with mouse and switch to touch works, and after that only touch works.

zefj:
The issue seems to be caused by the [multitouch](https://github.com/Popmotion/popmotion/blob/master/packages/popmotion/src/input/multitouch/index.ts) function, which does not reset `isTouchDevice` flag after the interaction finishes. I hot-wired the module and added the reset to the `stop` function like so:

[packages/popmotion/src/input/multitouch/index.ts#L70](https://github.com/Popmotion/popmotion/blob/master/packages/popmotion/src/input/multitouch/index.ts#L70)
```
  return {
    stop: () => {
      cancelOnFrameUpdate(updatePoint);
      updateOnMove.stop();
      isTouchDevice = false;
    }
  };
```

which works well for me. This is a result of a quick debug, I am not familiar with the internals of the project, so could you (@InventingWithMonster) please look at it? I'll be happy to prepare a PR if that fixes the problem without any side effects.

voxpelli:
I think the real issue here are the API:s that's used. I've created a new Popmotion module that's created around the unified Pointer Events API (and which works with polyfills). It's currently successfully deployed on two large Swedish web sites.

InventingWithMonster:
Closing this as rethinking events

PR #444, opened 2018-08-27
Set if isTouchDevice

joshuacrowley:
Fix for devices with touchscreens and mouse input. As detailed in #265  currently once an element is dragged, the mouse input stops working.

### DIFF
--- a/packages/popmotion/src/input/multitouch/index.ts
+++ b/packages/popmotion/src/input/multitouch/index.ts
@@ -79,6 +79,7 @@ const multitouch = ({
       stop: () => {
         cancelOnFrameUpdate(updatePoint);
         updateOnMove.stop();
+        isTouchDevice = false;
       }
     };
   });


joshuacrowley:
This is my first open source PR, so perhaps missed a step here. Let me know if you need something else.

Andarist:
I have no idea if the change itself is correct but at the very least you should rename this variable now.

voxpelli:
While this solution may perhaps mitigate some issues, does not solve the core issue of multiple types of inputs – only the Pointer Events API does that – see my response in #265. I will open source the module I created at @Sydsvenskan soon, when I'm fully back from vacation.

InventingWithMonster:
Closing this as it won't fix the underlying issue. I'll have to revisit this at some point, there is an open issue here https://github.com/Popmotion/popmotion/issues/265

PR #452, opened 2018-08-30
Adding support for `prefers-reduced-motion` media query

mattgperry:
This currently sets animations to their end state immediately if a user's device reports `prefers-reduce-motion`

Currently blocked until we can figure out a way to differentiate direct input poses from animations (this currently breaks dragging)


packages/popmotion-pose/src/index.ts
```diff
@@ -1,3 +1,5 @@
 import domPose from './dom/pose';
+import { setReducedMotion } from './inc/accessibility';
```
Andarist: u can just do:
```js
export { default } from './dom/pose';
export { setReducedMotion } from './inc/accessibility';
```

InventingWithMonster: Ah that's nice thanks though I've taken it out for now in favour of forcing animations in 'development'

packages/popmotion-pose/src/inc/accessibility.ts
```diff
@@ -0,0 +1,17 @@
+let prefersReducedMotion = false;
+
+if (typeof window !== 'undefined') {
+  const motionMediaQuery = window.matchMedia('(prefers-reduced-motion)');
+
+  const setReducedMotionPreferences = () => {
+    prefersReducedMotion =
+      motionMediaQuery.media === 'not all'
+        ? false
+        : motionMediaQuery.matches ? true : false;
```
Andarist: `motionMediaQuery.matches` is already a boolean so this is literally casting a boolean to boolean, can be shortened to
```js
prefersReducedMotion = motionMediaQuery.media === 'not all'
  ? false
  : motionMediaQuery.matches
```

Andarist: also - cant find the justification for special handling `"not all"`, i might be missing something though, in all my tests when this condition is met `motionMediaQuery.media === 'not all'` then this is also true `motionMediaQuery.matches === false`, suggesting u could use just a simple
```js
prefersReducedMotion = motionMediaQuery.matches
```

InventingWithMonster: Yeah excellent points

packages/pose-core/src/factories/setter.ts
```diff
@@ -157,13 +160,15 @@ const createPoseSetter = <V, A, C, P>(
...
-    convertTransitionDefinition
+    convertTransitionDefinition,
+    isReducedMotion
   } = setterProps;
 
   return (next: string, nextProps: Props = {}, propagate: boolean = true) => {
     const { children, values, props, activeActions, activePoses } = state;
     const { delay = 0 } = nextProps;
     const hasChildren = children.size;
+    const useTransition = !isReducedMotion() || isDev;
```
Andarist: after thinking about it a little I don't think that forcing animation in dev is a good thing, it's good to test those things out during development too and it would be weird to have pose animations on and everything else off

I realize that somebody might be surprised about their animations being off when forgetting that they use reduced motion setting - but it seems to me that one time console.warn in dev mode when `isReducedMotion() === true` would be sufficient to help people realize this.

packages/popmotion-pose/src/inc/accessibility.ts
lifeiscontent: @InventingWithMonster just a side note, it might be nice to configure the prefersReducedMotion as an option since not all browsers implement it, we could instead make it a user preference in a UI somewhere.

lifeiscontent:
@InventingWithMonster @Andarist any updates on this feature?

packages/popmotion-pose/src/index.ts
lifeiscontent: I'd rather see this as an option to leave it up to a developer to choose when to disable/enable the animation

e.g. somewhere in a config set `disableAnimation: true` and have popmotion respect that rather than dealing with prefers reduce motion and browser inconsistencies between it, the code written in this library to support is will probably be similarly written within the app the user has written as well which just makes more code bloat in the resulting app.

green-arrow:
@InventingWithMonster - I'm curious about this feature as well. For visual regression testing with something like https://percy.io it would be awesome to be able to disable animations instead of relying on putting `wait`s in tests. Is this planned for the near future or is it something that will take a while to make it in?

PR #462, opened 2018-09-03
Fix dynamic positional unit conversion (#461)

### LINKED ISSUE #461: dynamic prop with random poseKey not updating underlying dom element attribute
Here is a demo showing the problem: https://codesandbox.io/s/186wz3zkq7.

I expect `style.top` to be getting updated continually. It appears to only take the value returned upon first invocation of the dynamic prop, however.

Note that the dynamic prop function is invoking as expected (see console.log output). The issue is in something after that value is returned.

~Probably I'm doing something wrong but maybe I'm hitting a bug, not sure. So far nothing that I've read in the docs makes clear to me that I've used the library wrong.~ Confirmed its a bug.

klaasman:
Positional props were not handled correctly when defined as a function. This fixes #461.

### DIFF
--- a/packages/popmotion-pose/src/dom/flip.ts
+++ b/packages/popmotion-pose/src/dom/flip.ts
@@ -7,6 +7,7 @@ import {
   PoserState,
   DomPopmotionPoser
 } from '../types';
+import { resolveProp } from './utils';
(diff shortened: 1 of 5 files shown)


klaasman:
You're welcome!

PR #470, opened 2018-09-05
Fixed resolving position when flipping

### LINKED ISSUE #353: FLIP does not transition from position 'relative' to 'fixed'
**Describe the bug**
Pose doesn't seem to like transitioning the `position` attribute. It remains `relative` even when a new pose is set with a `fixed` position.

As a side-effect, it also doesn't FLIP between 100px @ relative and 100% @ fixed.

**To Reproduce**
https://codesandbox.io/s/8zj05687z2

Click the second button, 'Pose'.

**Expected behavior**
I want it to switch instantly, so that FLIP can recalculate the element's position and do it's magic.

It should go fullscreen (as (3)Manual Pose does in the above codesandbox). Instead it remains inline.

**Desktop (please complete the following information):**
 - OS: macOS
 - Browser Chrome
 - Version 66.0.3359.181

Andarist:
Fixes #353

Alternatively maybe `'position'` should be a `positionalProp` too? Didn't want to make such a change though, because I'm not sure what unexpected consequences it could have.

### DIFF
--- a/packages/popmotion-pose/src/dom/flip.ts
+++ b/packages/popmotion-pose/src/dom/flip.ts
@@ -75,14 +75,14 @@ const explicitlyFlipPose = (state: PoserState, nextPose: Pose) => {
     ...remainingPose
   } = nextPose;
 
-  const propsToSet = positionalProps.reduce(
+  const propsToSet = positionalProps.concat('position').reduce(
     (acc, key) => {
       if (nextPose[key] !== undefined) {
         acc[key] = resolveProp(nextPose[key], state.props);
       }
       return acc;
     },
-    { position } as StyleMap
+    {} as StyleMap
   );
 
   elementStyler.set(propsToSet).render();


InventingWithMonster:
@Andarist Thanks for taking a look but I don't understand how this changes the behaviour? If `position: undefined` Stylefire won't set it

Andarist:
Well, I'm not aware of all the intricacies of the implementation, so it was kinda expected on my side that I might break something (unfortunately 😉).

The fix itself is fixing a "sub-issue" of #353. The original example used functions as values of `position`, and therefore they were not resolved at all and thus the animation didn't work. Demo of the problem might be found [here](https://codesandbox.io/s/j37p90pzkw) (this might be the same code - except deps versions - as the original one, aint sure, didnt compare them right now).

If `position: undefined` is needed I would say that this should fix it:
```js
  const propsToSet = positionalProps.concat('position').reduce(
    (acc, key) => {
      if (nextPose[key] !== undefined) {
        acc[key] = resolveProp(nextPose[key], state.props);
      }
      return acc;
    },
    { position } as StyleMap
  );
```
So the existing position has a chance to override itself (after being resolved - or not) & passing `position: undefined` would get preserved.

If you agree on this I'll amend the PR later.

InventingWithMonster:
Ah I see no I get it now, nice catch

InventingWithMonster:
Published as `3.2.7`

Andarist:
Just to reassure myself - `position: undefined` doesnt have to actually be passed as one of `propsToSet` in `elementStyler.set(propsToSet).render();` call?

InventingWithMonster:
No it’s irrelevant. I missed the bit where this fix passes it to resolveProp

PR #474, opened 2018-09-10
Support wrapping elements which use React.forwardRef

Andarist:
This should handle i.e. upcoming styled-component@4

### DIFF
--- a/packages/react-pose/src/components/PoseElement.tsx
+++ b/packages/react-pose/src/components/PoseElement.tsx
@@ -1,4 +1,5 @@
 import * as React from 'react';
+import * as ReactIs from 'react-is';
 import { createContext, createElement } from 'react';
 import poseFactory, {
   DomPopmotionPoser,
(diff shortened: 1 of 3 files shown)


packages/react-pose/src/components/PoseElement.tsx
```diff
@@ -1,4 +1,5 @@
 import * as React from 'react';
+import * as ReactIs from 'react-is';
```
InventingWithMonster: would `import { isForwardRef }` have any effect on tree-shaking?

Andarist: No, for 2 reasons:
- both flavours are equivalent until u actually try to use `ReactIs` as "object" (i think), i.e. if u do `someCall(ReactIs)` or `ReactIs['dynamic access']`, bundlers should be easily capable of recognizing this `import * as ReactIs` as merely a group of live bindings
- `react-is` is unfortunately not exposing `esm` build - which it actually could (I might prepare a PR for this), so i.e. webpack will just bail out from tree-shaking this as its a cjs module. Anyway - the module itself is rather small and probably will become common~ in the community, so it shouldn't matter much, as it has potential of being used by more libraries than only popmotion.

Andarist:
Done

InventingWithMonster:
I think it has to be a new PR, worked for your other one. Not to worry though, looks good. Thanks!

PR #477, opened 2018-09-11
Support a resolver function as dragBounds

Andarist:
This will require a typings update, but I'd like to know first if you'd like to incorporate this change or not before adjusting them.


Andarist:
I've edited typings really quickly, hope they are going to work (Im a TS noob unfortunately). If not I might try to fight them later.

Not sure about `ResolverProps` name, maybe you'd have something better up your sleeve.

PR #488, opened 2018-09-14
Remove surrounding backticks from some docs headers - they screwed up…

Andarist:
… element IDs generation

I believe those backticks caused generation of weird IDs and thus such corresponding links:
https://popmotion.io/pose/api/posegroup/#{{0}}-import"

With this change they should look like this:
https://popmotion.io/pose/api/posegroup/#posegroup-import"


InventingWithMonster:
Nice one thanks, this has been a pain

PR #490, opened 2018-09-16
Remove flip from poseValues

Andarist:


InventingWithMonster:
Is there a point you've seen this property slipping through? `pose-core` technically shouldn't really know about `flip`, as its a DOM-specific prop that `popmotion-pose` shouldn't be passing forward

Andarist:
yeah, I've seen it slipping to `createValueMap` - https://github.com/Popmotion/popmotion/blob/066c11e000b300f0774ff82527752a00fe571197/packages/pose-core/src/factories/values.ts#L214 and thus becoming a Value unnecessarily

If you know at which level this should be fixed, let me know and I'll adjust the PR

InventingWithMonster:
Ah shit yeah makes sense, in `popmotion-pose/src/dom/pose.ts` line 83
there's a function `transformPose` that returns poses without `flip` - it's
become way more side-effecty than it was originally designed.

Maybe there's scope to add a new optional config function to `pose-core`
called `getValuesFromPose` that in `popmotion-pose` `({ flip, ...pose }) =>
pose` and is undefined in `animated-pose`? What do you think?

Andarist:
I still don't have big picture here, don't always see what are responsibilities for various packages. I'm just playing around, debug often and when I see something off I report it back or try to fix :p 

Could you tell me more about `transformPose` logic? I've also noticed that in general when I define visible & hidden poses there is additional pose created - called `flip`. Is that intentional or is it also some kind of leftover artifact?

Could `flip` be normalized during initialization and removed from the config? That would prevent it being passed around as its part, not sure how internals work exactly, so its hard to judge this for me.

As to `getValuesFromPose` - it depends how many platform-specific things there are gonna be, is it supported to create your own `pose-core` flavours? Extra configuration makes things more complex, so I'm wondering if there is a way to avoid it.

InventingWithMonster:
Generally the core Pose logic (in `pose-core`) should be independent of
both animation library and renderer.

So for instance the `popmotion-pose` package itself has two factories, one
handles Popmotion-specific code and one handles an intergration between
that and the DOM. There might be another one day that integrates Popmotion
Pose with Three.js (for instance)

I might look into a class-based architecture but I'm not convinced it'll
make anything clearer. The abstractions themselves make things complicated
but it has also allowed me to consolidate a lot of logic and for instance
launch Pose for Vue with feature-parity four days after reading Vue docs
for the first time. It's a double-edged sword for sure.

`transformPose` itself accepts a pose that might have data specific to that
pose flavour and returns it in a format that Pose core should care about,
which is essentially just responsible for deciding when values are created
and animated. FLIP only really makes sense in a DOM-context. I was
experimenting with something specific to React Native but it'd be done in a
totally different way.

So `transformPose` is an opportunity for the `popmotion-pose` DOM
integration to say to core Pose "the user says you want to be animating
width and height but actually you want to be animating x and y like this"
but for it to know what x and y are it needs to actually do some DOM
measurements.

The flip pose works because the `flip` method exported from
`popmotion-pose` does `poser.set('flip')` and the pose core can say yeah we
do have a pose by that name. A cleaner and more flexibile way would be to
allow `set` to accept arbitrary poses that haven't been defined on
initialisation, but then we'd need to bake in a whole thing for a poser to
flag when new animated values have been added so (for instance) Animated
Pose could kick off a React Native re-render

Andarist:
> I might look into a class-based architecture but I'm not convinced it'll make anything clearer.

I like the composition approach used, I just havent found time yet to understand this hierarchy, that's why I was asking.

So - based on what you have explained to me. `transformPose` would be good place to filter out `flip`, right? 

As to the flip pose - I guess I need to read more through the code to see how flipping works now to understand why `poser.set('flip')` is useful. Seems to me like it shouldnt be a pose, but rather only a characteristic to describe how to transition from pose A to pose B

InventingWithMonster:
“Problem” in this specific instance is transformPose doesn’t get run when
we scrape the poses for values we’re going to animate. In a way we won’t
even know which values we need to animate to FLIP between two positions
before we measure the two states.

Andarist:
Ok, that makes it harder then. Not sure what the best solution is, not being that confident with the codebase. I'll let it sit for now, maybe you'll figure out how you would like to handle this ideally. If you come up with the preferred solution I may tweak my PR if you let me know.

InventingWithMonster:
Decided this one isn't going to hurt but we'll keep an eye out for the problem reoccuring

PR #492, opened 2018-09-17
PoseGroup props forwarding

InventingWithMonster:


packages/react-pose/src/components/Transition/children.ts
```diff
@@ -79,17 +74,18 @@ const mergeChildren = ({
     const newChild = cloneElement(child, {
       _pose: exitPose,
       onPoseComplete: removeFromTree(key),
-      popFromFlow: flipMove
+      popFromFlow: flipMove,
+      ...childProps
```
Andarist: looking at this part of the algo it seems to me that props comming from group/Transition (that are represented here by childProps) override the actual used child props, are you sure this is expected behaviour?

PR #496, opened 2018-09-18
filtering previously set props

### LINKED ISSUE #494: React pose: Transition props forwarding bug
## Describe the bug

Props resolvers get stale props.

Stale values are passed here - https://github.com/Popmotion/popmotion/blob/009e777b90100c08e9dae16564babae14682eb2f/packages/react-pose/src/components/Transition/children.ts#L88-L89

This happens because we mutate the children, so when the child enters it receives `propsForChild` as its own props and later when the same child exits its now own props (received previously from `propsForChild`) override current `propsForChild` thus resulting in stale values received. 

## How to reproduce
Write a simple exit/enter animation with resolved value. Enter the item, remove it - observe the resolver calls.

## Expected behaviour
Current values being passed to resolvers.

InventingWithMonster:
Fixes #494

### DIFF
--- a/packages/react-pose/src/components/Transition/children.ts
+++ b/packages/react-pose/src/components/Transition/children.ts
@@ -2,11 +2,15 @@ import * as React from 'react';
 import { ReactElement } from 'react';
 import { CurrentPose } from '../PoseElement/types';
 import { MergeChildrenProps, Props } from './types';
+import { filterBy } from '../../utils/object-assign';
 const { Children, cloneElement } = React;
 
 const getKey = (child: ReactElement<any>): string => child.key as string;
 
-const filterChildProps = ({ children, _pose, onPoseComplete, popFromFlow, ...props }: Props) => props;
+const filterChildProps = (
+  { children, _pose, onPoseComplete, popFromFlow, ...props }: Props,
+  filterKeys?: string[]
+) => (filterKeys ? filterBy(props, filterKeys) : props);
 
 const animateChildrenList = (
   incomingChildren: Array<ReactElement<any>>,
@@ -68,11 +72,14 @@ const mergeChildren = ({
     };
 
     if (entering.has(child.key as string)) {
+      // If child is entering
       newChildProps.initialPose = preEnterPose;
-      newChildProps._pose = enterPose;
+      newChildProps._pose = enterPose; // TODO: Remove _pose and merge with child.props.pose
     } else if (moving.has(child.key as string) && flipMove) {
+      // If child is moving and we're using `flip`
       newChildProps._pose = [enterPose, 'flip'];
     } else {
+      // If child is moving
       newChildProps._pose = enterPose;
     }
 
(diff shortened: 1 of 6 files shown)


packages/react-pose/src/components/Transition/children.ts
```diff
@@ -68,11 +72,14 @@ const mergeChildren = ({
     };
 
     if (entering.has(child.key as string)) {
+      // If child is entering
       newChildProps.initialPose = preEnterPose;
-      newChildProps._pose = enterPose;
+      newChildProps._pose = enterPose; // TODO: Remove _pose and merge with child.props.pose
```
Andarist: sideways note - would it be possible to give `_pose` a better name? :p just by reading it I have no idea how it's used

packages/react-pose/src/components/Transition/children.ts
```diff
@@ -68,11 +72,14 @@ const mergeChildren = ({
...
     } else if (moving.has(child.key as string) && flipMove) {
+      // If child is moving and we're using `flip`
       newChildProps._pose = [enterPose, 'flip'];
     } else {
+      // If child is moving
```
Andarist: previous condition checks against `moving.has(child.key as string)` to test if the child is moving, this branch of the code is unconditional, is it intended?

packages/react-pose/src/components/Transition/children.ts
```diff
@@ -86,7 +93,7 @@ const mergeChildren = ({
       onPoseComplete: removeFromTree(key),
       popFromFlow: flipMove,
       ...propsForChild,
-      ...filterChildProps(child.props)
+      ...filterChildProps(child.props, Object.keys(propsForChild))
```
Andarist: this approach is IMHO faulty for edge cases, it depends though what kind of edge cases do you want to support

a possible situation is that we first render a child with no prop A inside a Transition with that prop A, the prop gets merged onto this child, later we rerender and the child receives a prop A this time (totally possible, although imho a code smell), the new child prop A should "win" but instead you filter it out here

also the problem here might be that `propsForChild` are different between rerenders and this might lead to buggy behaviours

in this exact scenario my first point is not so valid because this branch handles only leaving nodes, but I'm wondering if the same bug doesn't also affect any other situation? doesn't it also affect moving elements?

Andarist:
also a downside of this approach is that we do a lot of filtering, copying etc which adds up and decreases the perf (sure, might not matter much at the moment, but there has to be some threshold when it starts to matter)

packages/react-pose/src/components/Transition/children.ts
InventingWithMonster: It’s basically just a private pose that I added cuz I didn’t know how to access child props

packages/react-pose/src/components/Transition/children.ts
InventingWithMonster: It checks for that and if flipMove is true, so this does get hit if flipMove is false

Andarist: but this last branch doesnt check against `moving.has(child.key as string)` so I'm this comment looks off to me, the last `else` branch is "activated" when:
- `moving.has(child.key as string) && !flipMove`
- `!moving.has(child.key as string) && flipMove`
- `!moving.has(child.key as string) && !flipMove`

it covers 3 supplementary cases (in relation to the previous condition) and in 2 of them the child is NOT moving

PR #500, opened 2018-09-18
Validate <Transition/> children in dev builds

Andarist:


packages/react-pose/src/components/Transition/children.ts
```diff
@@ -122,13 +123,35 @@ export const handleIncomingChildren = (props: MergeChildrenProps) => {
...
+      }
+
+      warning(
+        child.type !== 'string',
+        '<Transition /> (or <PoseGroup/>) children have to be composite components.\n'
+        'Instead it received: ${child.type}.'
+      );
+
+      if (process.env.NODE_ENV !== 'production') {
+        if (list.length === 1) {
```
InventingWithMonster: I don't understand when this would catch what the following warning wouldn't?

Andarist: a single child, I assume key-less single child is allowed, or isnt it? if it isnt then ofc this can be removed

InventingWithMonster: Yeah all children must have a key otherwise we don't know if a child has been switched out

Andarist:
Closing this, because in the meantime you have already implemented same validation in `getKey` in the other PR.

PR #501, opened 2018-09-18
Alternative fix/override child props

Andarist:
**I have not tested this yet at all**, please review the code/approach itself - hopefully also tests will be able to validate my logic here.

Additional benefit of this approach is that children during initial mount are also receiving group props now - which was not true before, only merged children (so subsequent renders) received them.

### DIFF
--- a/packages/react-pose-core/src/components/PoseComponent.tsx
+++ b/packages/react-pose-core/src/components/PoseComponent.tsx
@@ -89,7 +89,7 @@ class PoseComponent extends React.PureComponent<PoseComponentProps> {
     const poseList: string[] = Array.isArray(pose) ? pose : [pose];
 
     Promise.all(poseList.map(key => key && this.poser.set(key))).then(
-      () => onPoseComplete && onPoseComplete()
+      () => onPoseComplete && onPoseComplete(pose)
     );
   }
 
--- a/packages/react-pose/src/components/PoseElement/index.tsx
+++ b/packages/react-pose/src/components/PoseElement/index.tsx
@@ -15,7 +15,7 @@ import {
 import isValidProp from '@emotion/is-prop-valid';
 import { invariant } from 'hey-listen';
 import { hasChanged } from '../../utils/has-changed';
-import { pickAssign } from '../../utils/pick-assign';
+import { pickAssign } from '../../utils/object-assign';
 
 export const PoseParentContext = createContext({});
 
--- a/packages/react-pose/src/components/PoseElement/index.tsx
+++ b/packages/react-pose/src/components/PoseElement/index.tsx
@@ -280,7 +280,7 @@ class PoseElement extends React.PureComponent<PoseElementInternalProps> {
     const poseList: string[] = Array.isArray(pose) ? pose : [pose];
 
     Promise.all(poseList.map(key => key && this.poser.set(key))).then(
-      () => onPoseComplete && onPoseComplete()
+      () => onPoseComplete && onPoseComplete(pose)
     );
   }
 
--- a/packages/react-pose/src/components/PoseElement/types.ts
+++ b/packages/react-pose/src/components/PoseElement/types.ts
@@ -23,6 +23,7 @@ export type PoseElementProps = {
   _pose?: CurrentPose;
   initialPose?: CurrentPose;
   withParent?: boolean;
+  onPoseComplete?: (pose: CurrentPose) => any;
   onValueChange?: { [key: string]: (v: any) => any };
   innerRef?: { current: any } | RefFunc;
   [key: string]: any;
--- a/packages/react-pose/src/components/Transition/children.ts
+++ b/packages/react-pose/src/components/Transition/children.ts
@@ -21,22 +21,20 @@ const animateChildrenList = (
 
 const mergeChildren = ({
   incomingChildren,
-  displayedChildren,
+  children,
   isLeaving,
   removeFromTree,
   groupProps
 }: MergeChildrenProps) => {
   const {
-    children: groupChildren,
     preEnterPose,
     enterPose,
     exitPose,
     flipMove,
-    animateOnMount,
   } = groupProps;
-  const children: Array<ReactElement<any>> = [];
+  const mergedChildren: Array<ReactElement<any>> = [];
 
-  const prevKeys = displayedChildren.map(getKey);
+  const prevKeys = children.map(getKey);
   const nextKeys = incomingChildren.map(getKey);
 
   const entering = new Set(
@@ -73,11 +71,11 @@ const mergeChildren = ({
       newChildProps._pose = enterPose;
     }
 
-    children.push(cloneElement(child, newChildProps));
+    mergedChildren.push(cloneElement(child, newChildProps));
   });
 
   leaving.forEach(key => {
-    const child = displayedChildren.find(c => c.key === key);
+    const child = children.find(c => c.key === key);
     const removeChildFromTree = removeFromTree(key)
 
     const newChild = cloneElement(child, {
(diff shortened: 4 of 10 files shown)


Andarist:
Ok, so I've fixed tests 🚀 

Other than the whole group/child props merging this PR addresses few other issues:
- doesnt override child's `onPoseComplete` with `Transition`'s `onPoseComplete` - additionally I've added a dev-only warning when somebody uses `onPoseComplete` for `Transition`, IMHO it doesn't make sense on it, some other callback might be useful though (it's being discussed in other issues)
- `onPoseComplete` gets called with `pose` argument
- `onPoseComplete` prop got typed
- `propsForChild` are also passed to childs during initial mount, in the previous version of the code there were only passed when merging children
- `onPoseComplete` gets called for leaving children

If you decide to move with this version of the code I will add tests for some of the cases mentioned above

packages/react-pose/src/components/Transition/children.ts
```diff
@@ -62,31 +57,36 @@ const mergeChildren = ({
...
+    const child = children.find(c => c.key === key);
+    const removeChildFromTree = removeFromTree(key)
+
     const newChild = cloneElement(child, {
       _pose: exitPose,
-      onPoseComplete: removeFromTree(key),
+      onPoseComplete: pose => {
+        removeChildFromTree()
+        const { onPoseComplete } = child.props
+        // TODO: call original onPoseComplete prop also in react-pose-core, vue-pose
```
Andarist: probably should be done as part of this PR - but didn't want to touch those without your approval first

Andarist:
Also - after thinking about this problem a little bit while developing this version of the fix I feel like it would be better NOT TO provide `propsForChild` but rather `groupProps` (although right now when im writing it it seems that we would have to consider shallow equality check for them to bring back PureComponent semantics, as passing groupProps "as is" to the child would totally make `PoseElement` being a PureComponent worthless as obviously object props are not easily comparable) and just `groupProps` with `child.props` when calling resolvers. It's slightly surprising to me that the current solution allows `groupsProps` to be accessed inside the child itself (render, methods etc)

packages/react-pose-core/src/components/PoseComponent.tsx
```diff
@@ -89,7 +89,7 @@ class PoseComponent extends React.PureComponent<PoseComponentProps> {
     const poseList: string[] = Array.isArray(pose) ? pose : [pose];
 
     Promise.all(poseList.map(key => key && this.poser.set(key))).then(
-      () => onPoseComplete && onPoseComplete()
+      () => onPoseComplete && onPoseComplete(pose)
```
InventingWithMonster: Very nice idea! Would it be better here to pass `key`? That way its simpler in `onPoseComplete` to compare against a specific pose

packages/react-pose/src/components/PoseElement/index.tsx
```diff
@@ -280,7 +280,7 @@ class PoseElement extends React.PureComponent<PoseElementInternalProps> {
     const poseList: string[] = Array.isArray(pose) ? pose : [pose];
 
     Promise.all(poseList.map(key => key && this.poser.set(key))).then(
-      () => onPoseComplete && onPoseComplete()
+      () => onPoseComplete && onPoseComplete(pose)
```
InventingWithMonster: As above

packages/react-pose/src/components/PoseElement/types.ts
```diff
@@ -23,6 +23,7 @@ export type PoseElementProps = {
   _pose?: CurrentPose;
   initialPose?: CurrentPose;
   withParent?: boolean;
+  onPoseComplete?: (pose: CurrentPose) => any;
```
InventingWithMonster: This would be `pose: string`

packages/react-pose/src/components/Transition/children.ts
```diff
@@ -6,8 +6,6 @@ const { Children, cloneElement } = React;
 
 const getKey = (child: ReactElement<any>): string => child.key as string;
 
-const filterChildProps = ({ children, _pose, onPoseComplete, popFromFlow, ...props }: Props) => props;
```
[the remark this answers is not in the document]
InventingWithMonster: It was to ensure we weren't passing props that would affect the child component if we weren't intending to. If this PR fixes that without it, all the better

packages/react-pose/src/components/Transition/children.ts
InventingWithMonster: Yeah great idea

packages/react-pose/src/components/Transition/index.tsx
```diff
@@ -19,17 +20,39 @@ class Transition extends React.Component<Props, State> {
...
+      flipMove,
+      animateOnMount,
+      ...propsForChild,
+    } = props;
 
-    return {
+    const incomingChildren = makeChildList(groupChildren);
+
+    if (process.env.NODE_ENV !== 'production') {
+      warning(!propsForChild.onPoseComplete, '<Transition/> (or <PoseGroup/>) doesn\'t accept onPoseComplete prop.')
```
InventingWithMonster: Just a style thing but I'd prefer backticks here without string escaping. Also although we had a ticket about it I wonder about adding warnings about unsupported props. It's a rabbit hole that we could easily litter the codebase with and there was no documentation to suggest it is supported

packages/react-pose/src/components/Transition/index.tsx
```diff
@@ -19,17 +20,39 @@ class Transition extends React.Component<Props, State> {
...
+      exitPose,
+      flipMove,
+      animateOnMount,
+      ...propsForChild,
+    } = props;
 
-    return {
+    const incomingChildren = makeChildList(groupChildren);
+
+    if (process.env.NODE_ENV !== 'production') {
```
InventingWithMonster: Will this be correctly transpiled/minified away in prod?

packages/react-pose/src/utils/object-assign.ts
```diff
@@ -10,3 +10,14 @@ export const pickAssign = (shouldPick: TestString, sources: Props[]) =>
     }
     return picked;
   }, {});
+
+export const filterBy = (obj: Props, keys: string[]) =>
```
InventingWithMonster: I don't think this is used in this PR anymore?

InventingWithMonster:
Looks awesome man thanks, just got some comments/questions/couple changes.

packages/react-pose/src/components/Transition/children.ts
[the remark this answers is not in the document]
Andarist: hm, that was a function that filtered out `child.props` itself to omit those, the only thing I can think of is that Transition would like to override those and then we wouldnt like to override those coming from Transition with child.props, but I can't see Transition wanting to do that, so I'm totally unsure what was the purpose of this function - I might be missing something though

packages/react-pose-core/src/components/PoseComponent.tsx
Andarist: there might be multiple keys (see `Promise.all` usage), so we have to pass either `pose` or `poseList` here

packages/react-pose/src/components/PoseElement/types.ts
Andarist: if we change it, currently the typing is correct (right?) - `CurrentPose` might be an array of strings

packages/react-pose/src/components/Transition/index.tsx
Andarist: This is valid concern, I've added it for 2 reasons though
- you've tried to pass it in previous version of the fix [here](https://github.com/Popmotion/popmotion/pull/496/files#diff-d6d4f941ac5e823f38de0ba45db686a8R255) 😄 
- but on a serious note - `<Transition onPoseComplete={() => {}}>` would override child's `onPoseComplete` and while learning the API it might be an easy thing to miss, so to improve DX here I would advocate to warn about this case, especially that this is dev-only warning so won't cost anything in production builds

packages/react-pose/src/components/Transition/index.tsx
Andarist: yeah

packages/react-pose-core/src/components/PoseComponent.tsx
InventingWithMonster: Ah of course I misread my own code

packages/react-pose/src/components/PoseElement/types.ts
InventingWithMonster: I meant in the event of changing the above, but yeah this is fine

packages/react-pose/src/components/Transition/index.tsx
InventingWithMonster: Yeah makes sense, good thinking

PR #503, opened 2018-09-19
Framesync 4

InventingWithMonster:


packages/framesync/.gitignore
```diff
@@ -0,0 +1,61 @@
+# Logs
```
Andarist: this is covered by root `.gitignore`, u should be able to remove this entire file

PR #509, opened 2018-09-21
Batching Transition state changes and add `onRest`

### LINKED ISSUE #363: React Pose: Performance improvement in PoseGroup
Demo: https://codepen.io/anon/pen/mKEzVp?editors=0010

Currently, when `PoseGroup` triggers leave animations, it removes the animating element from the DOM once that animation is finished.

The problem is that when lots of other animations are still animating, this React reconciliation stutters.

This ticket proposes that we wait until all known animations finish before removing children DOM nodes, which we will do in one go.

A serious concern is what happens if a second state change occurs while the current DOM elements are still being animated? It might become difficult to know which to remove.

There's a weird side effect where after this stuttering (yellow), activity seems to stop before picking up again:

image

Visually this looks like a pause in the animation. This performance improvement might fix this alone, but if it doesn't we'll need to investigate this also

### LINKED ISSUE #504: React Native <Transition />: cannot read property 'key' of null
## Ensure you're using the latest version
```
{
  "dependencies": {
    "@emotion/native": "9.2.9",
    "react-native-pose": "0.8.0"
  }
}
```

## Add the package name to bug title
`react-native-pose`

## Describe the bug
When I try to create a [switching visibility](https://popmotion.io/pose/api/native-transition/#{{0}}-usage-switching-visibility) pose, it throws an error: `Cannot read property 'key' of null`.

## How to reproduce
1. Follow the [switching visibility](https://popmotion.io/pose/api/native-transition/#{{0}}-usage-switching-visibility) guide on the docs.

## Expected behaviour
To not throw error 😅

## Link to CodePen example (or similar)
https://snack.expo.io/@kevinwolf/popmotion-issue

## Device information
Expo Snack and local react native app.

InventingWithMonster:
Closes #504 
Closes #363

### DIFF
--- a/packages/popmotion-pose/docs/api/react-native/native-transition.md
+++ b/packages/popmotion-pose/docs/api/react-native/native-transition.md
@@ -125,6 +125,16 @@ If `true`, `Transition` will finish animating exiting components out **before** 
 
 This can be useful when animating two or more components that affect each other's layout, which can lead to snapping when the exiting components are removed.
 
+### onRest
+
+`onRest: Function`
+
+When a component finishes exiting, it isn't removed immediately. Instead, it's kept in the DOM until **all** currently leaving components have finished animating out to prevent expensive layout thrashing.
+
+`onRest` will fire when all exit transitions are complete.
+
+If new exit transitions begin in the meantime, `onRest` won't be fired until these have also finished.
+
 ### ...props
 
 All remaining props passed to `Transition` will be forwarded to its immediate children, for use in dynamic props.
--- a/packages/react-pose-core/src/components/Transition/index.tsx
+++ b/packages/react-pose-core/src/components/Transition/index.tsx
@@ -2,7 +2,7 @@ import * as React from 'react';
 import handleChildrenTransitions from './children';
 import { Props, State } from './types';
 
-class Transition extends React.PureComponent<Props> {
+class Transition extends React.Component<Props, State> {
   static defaultProps = {
     preEnterPose: 'exit',
     enterPose: 'enter',
@@ -23,12 +23,14 @@ class Transition extends React.PureComponent<Props> {
 
   removeChild(key: string) {
     const { children, leaving } = this.state;
-    const { enterAfterExit } = this.props;
+    const { enterAfterExit, onRest } = this.props;
     if (!leaving.hasOwnProperty(key)) return;
 
     leaving[key] = true;
 
-    if (!Object.keys(leaving).every(Boolean)) return;
+    if (!Object.keys(leaving).every(leavingKey => leaving[leavingKey])) {
+      return;
+    }
 
     const targetChildren = children.filter(
       child => !leaving.hasOwnProperty(child.key)
@@ -47,7 +49,7 @@ class Transition extends React.PureComponent<Props> {
           children: targetChildren
         };
 
-    this.setState(newState);
+    this.setState(newState, onRest);
   }
 
   render() {
--- a/packages/react-pose/src/components/PoseElement/index.tsx
+++ b/packages/react-pose/src/components/PoseElement/index.tsx
@@ -1,5 +1,5 @@
 import * as React from 'react';
-import { isForwardRef } from 'react-is';
+//import { isForwardRef } from 'react-is';
 import { createContext, createElement } from 'react';
 import poseFactory, {
   DomPopmotionPoser,
@@ -172,7 +172,8 @@ class PoseElement extends React.PureComponent<PoseElementInternalProps> {
     const refs: RefSetters = {};
     const { elementType } = this.props;
 
-    if (typeof elementType === 'string' || isForwardRef(elementType)) {
+    if (typeof elementType === 'string') {
+      //  || isForwardRef(elementType) TODO: Fix types for isForwardRef
       refs.ref = this.setRef;
     } else {
       refs.innerRef = this.setRef;
--- a/packages/react-pose/src/components/Transition/children.ts
+++ b/packages/react-pose/src/components/Transition/children.ts
@@ -34,6 +34,7 @@ const handleTransition = (
     animateOnMount,
     enterAfterExit,
     flipMove,
+    onRest,
     ...propsForChild
   }: Props,
   {
--- a/packages/react-pose/src/components/Transition/children.ts
+++ b/packages/react-pose/src/components/Transition/children.ts
@@ -1,58 +1,115 @@
 import * as React from 'react';
 import { ReactElement } from 'react';
-import { CurrentPose } from '../PoseElement/types';
-import { MergeChildrenProps, Props } from './types';
+//import { CurrentPose } from '../PoseElement/types';
+import { Props, State } from './types';
+import { invariant } from 'hey-listen';
 const { Children, cloneElement } = React;
 
-const getKey = (child: ReactElement<any>): string => child.key as string;
+const filterChildProps = ({
+  children,
+  _pose,
+  onPoseComplete,
+  popFromFlow,
+  ...props
+}: Props) => props;
 
-const filterChildProps = ({ children, _pose, onPoseComplete, popFromFlow, ...props }: Props) => props;
+// const animateChildrenList = (
+//   incomingChildren: Array<ReactElement<any>>,
+//   pose: CurrentPose,
+//   initialPose?: CurrentPose
+// ) => {
+//   const children: Array<ReactElement<any>> = [];
 
-const animateChildrenList = (
-  incomingChildren: Array<ReactElement<any>>,
-  pose: CurrentPose,
-  initialPose?: CurrentPose
-) => {
-  const children: Array<ReactElement<any>> = [];
+//   Children.forEach(incomingChildren, (child: ReactElement<any>) =>
+//     children.push(cloneElement(child, { pose, initialPose }))
+//   );
+//   return children;
+// };
 
-  Children.forEach(incomingChildren, (child: ReactElement<any>) =>
-    children.push(cloneElement(child, { pose, initialPose }))
+// const mergeChildren = ({
+//   incomingChildren,
+//   displayedChildren,
+//   finishedLeaving,
+//   scheduleChildRemoval,
+//   groupProps
+// }: MergeChildrenProps) =>
+
+// export const handleIncomingChildren = (props: MergeChildrenProps) => {
+//   const { displayedChildren, incomingChildren, groupProps } = props;
+//   const { animateOnMount, preEnterPose, enterPose } = groupProps;
+
+//   // If initial mount and we're animating
+//   if (!displayedChildren && animateOnMount) {
+//     return animateChildrenList(incomingChildren, enterPose, preEnterPose);
+
+//     // If subsequent render
+//   } else if (displayedChildren) {
+//     return mergeChildren(props);
+
+//     // If initial mount and we're not animating
+//   } else {
+//     return animateChildrenList(incomingChildren, enterPose);
+//   }
+// };
+
+const getKey = (child: ReactElement<any>): string => {
+  invariant(
+    child && child.key !== null,
+    'Every child of Transition must be given a unique key'
   );
-  return children;
+
+  const childKey =
+    typeof child.key === 'number' ? child.key.toString() : child.key;
+
+  return childKey.replace('.$', '');
 };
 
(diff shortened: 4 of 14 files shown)


packages/popmotion-pose/docs/api/react-native/native-transition.md
```diff
@@ -125,6 +125,16 @@ If `true`, `Transition` will finish animating exiting components out **before**
 
 This can be useful when animating two or more components that affect each other's layout, which can lead to snapping when the exiting components are removed.
 
+### onRest
```
Andarist: Judging by the name I would also expect it to fire after all transition animations are complete - not necessarily only after leaving ones complete. What about entering animations? And moving?

packages/react-pose-core/src/components/Transition/index.tsx
```diff
@@ -2,7 +2,7 @@ import * as React from 'react';
 import handleChildrenTransitions from './children';
 import { Props, State } from './types';
 
-class Transition extends React.PureComponent<Props> {
+class Transition extends React.Component<Props, State> {
```
Andarist: good change 👌

packages/react-pose-core/src/components/Transition/index.tsx
```diff
@@ -23,12 +23,14 @@ class Transition extends React.PureComponent<Props> {
 
   removeChild(key: string) {
     const { children, leaving } = this.state;
-    const { enterAfterExit } = this.props;
+    const { enterAfterExit, onRest } = this.props;
     if (!leaving.hasOwnProperty(key)) return;
 
     leaving[key] = true;
```
Andarist: does this one has to be kept in state? its part of the state and u mutate it here, which is imho surprising in react's world, so maybe a class instance field would suffice for this? i generally try to stick to keeping in react's state only things related to render

packages/react-pose-core/src/components/Transition/index.tsx
```diff
@@ -23,12 +23,14 @@ class Transition extends React.PureComponent<Props> {
 
   removeChild(key: string) {
     const { children, leaving } = this.state;
-    const { enterAfterExit } = this.props;
+    const { enterAfterExit, onRest } = this.props;
     if (!leaving.hasOwnProperty(key)) return;
```
Andarist: is hasOwnProperty check needed here? like - wouldnt a "simpler" property lookup be enough?

packages/react-pose/src/components/PoseElement/index.tsx
```diff
@@ -172,7 +172,8 @@ class PoseElement extends React.PureComponent<PoseElementInternalProps> {
     const refs: RefSetters = {};
     const { elementType } = this.props;
 
-    if (typeof elementType === 'string' || isForwardRef(elementType)) {
+    if (typeof elementType === 'string') {
+      //  || isForwardRef(elementType) TODO: Fix types for isForwardRef
```
Andarist: technically this is a breaking change that got released as patch

packages/react-pose/src/components/Transition/children.ts
```diff
@@ -1,58 +1,69 @@
...
+  _pose,
+  onPoseComplete,
+  popFromFlow,
+  ...props
+}: Props) => props;
+
+const getKey = (child: ReactElement<any>): string => {
+  invariant(
+    child && child.key !== null,
+    'Every child of Transition must be given a unique key'
```
Andarist: is this true for a single child case?

packages/react-pose/src/components/Transition/children.ts
```diff
@@ -1,58 +1,69 @@
...
+  const nextKeys = targetChildren.map(getKey);
 
   const entering = new Set(
-    nextKeys.filter(key => isLeaving.has(key) || prevKeys.indexOf(key) === -1)
+    nextKeys.filter(
+      key => finishedLeaving.hasOwnProperty(key) || prevKeys.indexOf(key) === -1
+    )
   );
-  entering.forEach(key => isLeaving.delete(key));
+  entering.forEach(key => delete finishedLeaving[key]);
```
Andarist: im curious about the design here - `entering` is still a set, but `finishedLeaving` became a POJO, whats the reasoning behind this change?

packages/react-pose/src/components/Transition/children.ts
```diff
@@ -1,58 +1,69 @@
...
   const children: Array<ReactElement<any>> = [];
 
   const prevKeys = displayedChildren.map(getKey);
-  const nextKeys = incomingChildren.map(getKey);
+  const nextKeys = targetChildren.map(getKey);
 
   const entering = new Set(
-    nextKeys.filter(key => isLeaving.has(key) || prevKeys.indexOf(key) === -1)
+    nextKeys.filter(
+      key => finishedLeaving.hasOwnProperty(key) || prevKeys.indexOf(key) === -1
```
Andarist: `finishedLeaving.hasOwnProperty(key)` is used to "recycle" exited but not yet removed nodes?

packages/react-pose/src/components/Transition/children.ts
```diff
@@ -1,58 +1,69 @@
...
 
   const leaving = prevKeys.filter(
     key =>
-      !entering.has(key) && (isLeaving.has(key) || nextKeys.indexOf(key) === -1)
+      !entering.has(key) &&
+      (finishedLeaving.hasOwnProperty(key) || nextKeys.indexOf(key) === -1)
   );
 
-  leaving.forEach(key => isLeaving.add(key));
+  leaving.forEach(key => (finishedLeaving[key] = false));
```
Andarist: i think there might be a bug here - `leaving` might contain a key of a child that has already finished leaving (allowed because of the `finishedLeaving.hasOwnProperty(key)`) and u set it as not `finishedLeaving` here

packages/popmotion-pose/docs/api/react-native/native-transition.md
InventingWithMonster: Yeah it's a fair point and something I'd considered. However I thought that rerendering only when everything leaving has finished might not turn out to be good enough, and it could be the case we need to start waiting for entering animations too.

However there's a incongruence between the name and the function so I'll earmark it for reconsideration in the next major

packages/react-pose-core/src/components/Transition/index.tsx
InventingWithMonster: `getDerivedStateFromProps` is a static field so the only way to provide it scoped state is via `state`.

Mutating `leaving` is a bit naughty by the book (and we also provide a scoped function in state for the same reason that never gets updated) but if it's not done ignorant of the purpose of `state` and its immutability, there's nothing technically wrong with it.

packages/react-pose-core/src/components/Transition/index.tsx
InventingWithMonster: The reason I renamed this `finishedLeaving` in `react-pose` (and will do the same here at some point) is values in `finishedLeaving` can either be `true` or `false`.

`false` indicates it's leaving but hasn't finished, `true` it's finished.

So the alternative is `leaving[key] === undefined` which would be fine too, but also bikeshedding at this point

packages/react-pose/src/components/PoseElement/index.tsx
InventingWithMonster: IMO a breaking change is something that is documented API that changes, or you could consider it a bug as we technically supported it and then broke it rather than purposefully removed it. I don't even think it was released as it isn't in any patch notes.

I'll take a look at adding it back but there's no official typings and I was having trouble getting Rollup to respect my custom ones.

packages/react-pose/src/components/Transition/children.ts
InventingWithMonster: Yeah because if that single child switches out for another single child we need to know the difference.

packages/react-pose/src/components/Transition/children.ts
```diff
@@ -1,58 +1,69 @@
...
-) => {
-  const children: Array<ReactElement<any>> = [];
+  const childKey =
+    typeof child.key === 'number' ? child.key.toString() : child.key;
 
-  Children.forEach(incomingChildren, (child: ReactElement<any>) =>
-    children.push(cloneElement(child, { pose, initialPose }))
-  );
-  return children;
+  return childKey.replace('.$', '');
```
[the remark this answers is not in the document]
InventingWithMonster: It might be a holdover from React Native but over there the `key` sporadically seems to be `'.$'`, I haven't figured out why yet but seems to be after using `Children.isArray`

packages/react-pose/src/components/Transition/children.ts
InventingWithMonster: `finishedLeaving` could equally be a `Map` I guess, maybe it will be, but the difference is `finishedLeaving` holds boolean data whereas `entering` doesn't.

packages/react-pose/src/components/Transition/children.ts
InventingWithMonster: Here we're simply considering it entering if it's currently considered leaving

packages/react-pose/src/components/Transition/children.ts
InventingWithMonster: I'll look into it, thanks

packages/react-pose/src/components/Transition/children.ts
```diff
@@ -101,28 +114,15 @@ const mergeChildren = ({
...
-
-  // If initial mount and we're animating
-  if (!displayedChildren && animateOnMount) {
-    return animateChildrenList(incomingChildren, enterPose, preEnterPose);
-
-    // If subsequent render
-  } else if (displayedChildren) {
-    return mergeChildren(props);
+export default (props: Props, state: State) => ({
+  hasMounted: true,
```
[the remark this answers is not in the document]
InventingWithMonster: Maybe a different name like `hasInitialised` might be more correct, but it isn't unsafe for our purposes (detecting whether this is our first run or not)

packages/react-pose/src/components/Transition/index.tsx
```diff
@@ -1,54 +1,59 @@
...
-  handleIncomingChildren,
-  makeChildList,
-  removeFromChildren
-} from './children';
+import handleChildrenTransitions from './children';
 
 class Transition extends React.Component<Props, State> {
   static defaultProps = {
     flipMove: false,
+    enterAfterExit: false,
```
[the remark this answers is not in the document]
InventingWithMonster: I've yet to fully implement this

packages/react-pose-core/src/components/Transition/index.tsx
Andarist: Fair point about `leaving` having to be available in `getDerivedStateFromProps`. Maybe add a note/warning about leaving being part of the state but still being mutated here? Or don't destructure it out of `this.state` to make it more visible. 

Either way - logic is ok, so if you don't think a change is necessary then don't change anything :)

packages/react-pose/src/components/Transition/children.ts
Andarist: 👌

packages/react-pose/src/components/Transition/children.ts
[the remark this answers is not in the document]
Andarist: i like `hasInitialised` better, it's less confusing

packages/react-pose/src/components/Transition/children.ts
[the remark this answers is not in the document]
bwallberg: FYI https://reactjs.org/docs/react-api.html#reactchildren

> React.Children.toArray() changes keys to preserve the semantics of nested arrays when flattening lists of children. That is, toArray prefixes each key in the returned array so that each element’s key is scoped to the input array containing it.

You should probably document that you do this normalization somewhere - When I create a `<PoseGroup>` I map children that have gone through `toArray()` like so `<AnimatedDiv key={`wrapper-${child.key}`}>`  (AnimatedDiv being the pose component) and the above normalization broke my code, easy fix to simply remove '.$' from my key as well but still.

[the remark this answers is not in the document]
Andarist: I think I understand your issue, but could you prepare a simple codesandbox showcasing the problem? It would help me to fix it in react-pose itself and write a test that would prevent regressions.

PR #514, opened 2018-09-22
Avoid unnecessary props copying in <PoseGroup />, also make it inheri…

Andarist:
…t Component (instead of PureComponent)

### DIFF
--- a/packages/react-pose/src/components/Transition/PoseGroup.tsx
+++ b/packages/react-pose/src/components/Transition/PoseGroup.tsx
@@ -3,14 +3,12 @@ import { Props } from './types';
 import Transition from './';
 
 // TODO: Deprecate, add notice about Transition and flipMove: true opt-in
-export default class PoseGroup extends React.PureComponent<Props> {
+export default class PoseGroup extends React.Component<Props> {
   static defaultProps = {
     flipMove: true
   };
 
   render() {
-    const { children, ...props } = this.props;
-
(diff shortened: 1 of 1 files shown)


packages/react-pose/src/components/Transition/PoseGroup.tsx
```diff
@@ -3,14 +3,12 @@ import { Props } from './types';
...
+export default class PoseGroup extends React.Component<Props> {
   static defaultProps = {
     flipMove: true
   };
 
   render() {
-    const { children, ...props } = this.props;
-
-    return <Transition {...props}>{children}</Transition>;
+    return <Transition {...this.props} />;
```
Andarist: children is a prop, no need extract it from `this.props` by making a temporary/intermediate `props` object

packages/react-pose/src/components/Transition/PoseGroup.tsx
```diff
@@ -3,14 +3,12 @@ import { Props } from './types';
 import Transition from './';
 
 // TODO: Deprecate, add notice about Transition and flipMove: true opt-in
-export default class PoseGroup extends React.PureComponent<Props> {
+export default class PoseGroup extends React.Component<Props> {
```
Andarist: Same change as you have done to `Transition`, as its component that receives children (that won't evey be the same from reference equality stand of view) it's unnecessary here

InventingWithMonster:
Nice thanks

PR #515, opened 2018-09-22
Remove excessive window guard

Andarist:
migrated https://github.com/Popmotion/stylefire/pull/22

> No other `window` usage is guarded, also some other browser specific globals (like `HTMLElement`) are not guarded which tells me this is not intended to run in non-browser environment at all and thus this single check can be safely removed.


InventingWithMonster:
Sorry I forgot about these when I migrated (big part of the reason why I did!)

Andarist:
No problem 😃

PR #518, opened 2018-09-23
Fix issue with Transition and children props merging

### LINKED ISSUE #494: React pose: Transition props forwarding bug

Andarist:
closes #501 (its new version of the other PR), closes #496, fixes #494

**Note** I have not tested this code on any live example.

### DIFF
--- a/packages/react-pose/src/components/PoseElement/index.tsx
+++ b/packages/react-pose/src/components/PoseElement/index.tsx
@@ -82,23 +82,18 @@ class PoseElement extends React.PureComponent<PoseElementInternalProps> {
       typeof this.props.elementType === 'string' ? isValidProp : testAlwaysTrue;
   }
 
-  getInitialPose(): CurrentPose {
+  getInitialPose(): CurrentPose | void {
     const { getInitialPoseFromParent, pose, _pose, initialPose } = this.props;
 
     if (initialPose) {
       return initialPose;
     } else {
-      // Feels like this could probably be simpler
       const parentPose = getInitialPoseFromParent && getInitialPoseFromParent();
-      const thisPose = Array.isArray(pose) ? pose : [pose];
-      const thisInternalPose = Array.isArray(_pose) ? _pose : [_pose];
+      const initialPoses = (Array.isArray(parentPose) ? parentPose : [parentPose])
+        .concat(pose, _pose)
+        .filter(Boolean);
 
-      const initialPoses = Array.isArray(parentPose)
-        ? [...parentPose, ...thisPose, ...thisInternalPose]
-        : [parentPose, ...thisPose, ...thisInternalPose];
-
-      const filteredInitialPoses = initialPoses.filter(Boolean);
-      return filteredInitialPoses.length ? filteredInitialPoses : undefined;
+      return initialPoses.length > 0 ? initialPoses : undefined;
     }
   }
 
@@ -107,10 +102,11 @@ class PoseElement extends React.PureComponent<PoseElementInternalProps> {
     // If don't have an initial pose explicitly defined, return early
     if (!initialPose) return;
 
-    const thisPose = Array.isArray(pose) ? pose : [pose];
-    const thisInternalPose = Array.isArray(_pose) ? _pose : [_pose];
+    const firstPose = (Array.isArray(pose) ? pose : [pose])
+      .concat(_pose)
+      .filter(Boolean)
 
-    return [...thisPose, ...thisInternalPose];
+    return firstPose.length === 1 ? firstPose[0] : firstPose;
   }
 
   getSetProps() {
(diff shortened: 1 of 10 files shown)


packages/react-pose/src/components/PoseElement/index.tsx
```diff
@@ -82,23 +82,18 @@ class PoseElement extends React.PureComponent<PoseElementInternalProps> {
...
     const { getInitialPoseFromParent, pose, _pose, initialPose } = this.props;
 
     if (initialPose) {
       return initialPose;
     } else {
-      // Feels like this could probably be simpler
       const parentPose = getInitialPoseFromParent && getInitialPoseFromParent();
-      const thisPose = Array.isArray(pose) ? pose : [pose];
-      const thisInternalPose = Array.isArray(_pose) ? _pose : [_pose];
+      const initialPoses = (Array.isArray(parentPose) ? parentPose : [parentPose])
```
InventingWithMonster: 🎉🎉

packages/react-pose/src/components/Transition/children.ts
```diff
@@ -34,22 +36,35 @@ const handleTransition = (
     enterAfterExit,
     flipMove,
     onRest,
-    ...propsForChild
+    ...propsForChilds
```
InventingWithMonster: Should really be `propsForChild` (as they get applied per child) or `propsForChildren` which is the correct pluralisation. Either's fine by me

packages/react-pose/src/components/Transition/children.ts
```diff
@@ -34,22 +36,35 @@ const handleTransition = (
...
     finishedLeaving,
     hasInitialized,
+    indexedChildren: prevChildren,
     scheduleChildRemoval,
   }: State
 ) => {
   targetChildren = makeChildList(targetChildren);
 
-  const children: Array<ReactElement<any>> = [];
+  const nextState: Partial<State> = {
```
InventingWithMonster: TIL

Andarist:
I've added additional fix to this PR - please take a look at the last commit too. Issue demonstrated here - https://codesandbox.io/s/x3yk2jy3vo - notice that all children stay in the DOM after their exit animations finish.

Andarist:
Friendly ping :)

InventingWithMonster:
Ah yeah I'll take a look this morning thanks. Got a talk coming up and its
taking up most my time atm

InventingWithMonster:
If you go to `packages/react-pose` and run `yarn build` do you get type errors?

InventingWithMonster:
I'm going to add a step to circleCI that tests builds

InventingWithMonster:
Ok if you merge master back in that'll add these checks to the branch

Andarist:
I've rebased and I'm trying to fix the type errors, but my TS knowledge is really poor.

It's also weird that I'm getting type errors in the parts of the code that I haven't touched (I'm seeing several `Parameter 'paramName' implicitly has an 'any' type.` - they are quite easily fixable, but I have to annotate parts of the code that were not annotated previously, relaying on inference? not sure what TS can infer though).

Also most commonly from what I've seen things like `Props` (and maybe `State` but aint sure about this one) are typed as interfaces and you are using types for them, which are nominal, rather than structural from what I understand, therefore it's harder for me to type some code lines as TS is complaining about some things not being assignable to i.e. type State.

Anyway - gonna try to make it work, but would appreciate some insight into why things work how they work now (I'm especially curious about the whole implicit any thing, do you see those errors?)

InventingWithMonster:
I only see one error at a time with Rollup which is a bit whack-a-mole. I'm going to have a deeper look into it tomorrow. I'm not bound to using `type` for `Props` and `State`, if `interface` works better in this instance then we should use that.

Inference in TS is patchy imo. I tend towards it because its cleaner, but for instance it doesn't seem to recognise that `filter(Boolean)` will *never* return `void`, which is annoying.

Andarist:
It seems that I've managed to fix TS errors. Still unsure why I had problems at some places (mainly this implicit any thing) - I'll comment under changed lines, maybe you will know the reason behind the TS logic. If not - then screw it for now, need to finally read TS docs anyway, so maybe I'll find some answers there 😄 

Please review the last commit though to see if I havent done anything too stupid.

packages/react-pose/src/components/PoseElement/index.tsx
```diff
@@ -63,11 +63,11 @@ class PoseElement extends React.PureComponent<PoseElementInternalProps> {
    * =============================================
    */
   private childrenHandlers: PoseContextProps = {
-    registerChild: props => {
+    registerChild: (props: ChildRegistration) => {
```
Andarist: places like this surprise me, not sure why TS complains about this on my machine, but not on yours (and not on CI), looks like some config mismatch, but I shouldnt be using any extra config over what's in the repo here

InventingWithMonster:
LGTM thanks for your hard work on this one!

PR #521, opened 2018-09-24
Removed unnecessary Fragment usage (<Transition/> children have to ha…

Andarist:
…ve keys)


packages/react-pose/src/components/Transition/index.tsx
```diff
@@ -62,7 +61,7 @@ class Transition extends React.Component<Props, State> {
   }
 
   render() {
-    return <Fragment>{this.state.children}</Fragment>;
```
Andarist: arrays are accepted as render's return type, and as each child has to have a key the `Fragment` layer is not needed

Issue #528, opened 2018-09-28
Popmotion: transform.steps() doesn't round properly to each step. 


littlemilkstudio:
Love the library! This is just a small bug I ran into when I was testing my own project.

## Describe the bug
transform.steps() is rounding improperly to each step. Can be seen on boundaries between steps. Root cause stemming from stepProgress in calc.ts

## How to reproduce & Expected behavior
```
describe('stepProgress', () => {
  it('should return the correct progress 0 - 1 on border boundaries', () => {
    expect(stepProgress(3, 0)).toBe(0);
    expect(stepProgress(3, .24)).toBe(0);
    expect(stepProgress(3, .25)).toBe(.5);
    expect(stepProgress(3, .74)).toBe(.5);
    expect(stepProgress(3, .75)).toBe(1);
    expect(stepProgress(3, 1)).toBe(1);
  });
});
```

littlemilkstudio:
Suggested fix: [#529](https://github.com/Popmotion/popmotion/pull/529)

PR #529, opened 2018-09-28
Updated calc.stepProgress() to properly round around step border cases.

littlemilkstudio:
Fixes issue [#528](https://github.com/Popmotion/popmotion/issues/528).

### DIFF
--- a/packages/popmotion/src/_tests/calc.test.ts
+++ b/packages/popmotion/src/_tests/calc.test.ts
@@ -101,4 +101,13 @@ describe('stepProgress', () => {
     expect(stepProgress(3, .9)).toBe(1);
     expect(stepProgress(3, .1)).toBe(0);
   });
+  
+  it('should return the correct progress 0 - 1 on border boundaries', () => {
+    expect(stepProgress(3, 0)).toBe(0);
+    expect(stepProgress(3, .24)).toBe(0);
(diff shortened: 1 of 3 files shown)


InventingWithMonster:
Great contribution, thanks for this

packages/popmotion/src/_tests/calc.test.ts
```diff
@@ -101,4 +101,13 @@ describe('stepProgress', () => {
     expect(stepProgress(3, .9)).toBe(1);
     expect(stepProgress(3, .1)).toBe(0);
   });
+  
+  it('should return the correct progress 0 - 1 on border boundaries', () => {
```
Andarist: hm, is this correct behaviour? non edge segments are two times wider than edge ones, shouldnt distribution be even?

littlemilkstudio: My assumption was that each step would have an equal deviation from which it would snap to that step. kinda like a sinusoid where each step is a valley, once you get to a peak (odd numbered subsection) you'd go into the next valley. That assumption could be wrong tho.

My use case was achieving a scroll to this: http://www.benmingo.com/ . Would basically tell me the current project based on percentage of a container thats been scrolled.

Do you see it being used differently?

PR #531, opened 2018-09-29
Allow falsy values as <Transition/> children in TS typings

### LINKED ISSUE #341: Pose: TypeScript error with `PoseGroup` and single child
Ensure `PoseGroup` expects a single child as well as an array

Andarist:
fixes #341

Just reminding that I'm not a TS expert, but from my limited knowledge it seems like this should work.


DevanB:
Yes! Happy to see this might be fixed with the next release! 😄

DevanB:
@InventingWithMonster Any ETA when this fix might be released? I've had to turn off `strict` on my codebase until it is released (and would love to turn it back on!).

PR #533, opened 2018-09-30
Fix incorrect example (pressable -> focusable)

terrierscript:
I found incorrect doc.

Codesandbox sample is incorrect too but I can't fix it.
(If need codexandbox too, please close this issue)

### DIFF
--- a/packages/popmotion-pose/docs/examples/react/event-focus.md
+++ b/packages/popmotion-pose/docs/examples/react/event-focus.md
@@ -11,8 +11,19 @@ To animate an element on focus, set `focusable: true` and a `focus` prop.
 ```javascript
 const config = {
   focusable: true,
-  init: { scale: 1 },
-  press: { scale: 0.8 }
+  init: {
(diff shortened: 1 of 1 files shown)


Andarist:
rlly2kryrn is the codesandbox's ID of the focusable example

packages/popmotion-pose/docs/examples/react/event-focus.md
```diff
@@ -10,10 +10,10 @@ To animate an element on focus, set `focusable: true` and a `focus` prop.
 
 ```javascript
 const config = {
-  pressable: true,
+  focusable: true,
   init: { scale: 1 },
   press: { scale: 0.8 }
```
InventingWithMonster: Should be `focus`

terrierscript: Thank you!
I copy code from codesandbox

InventingWithMonster:
thanks!

PR #534, opened 2018-09-30
remove inline code mark for TOC

terrierscript:
Calcurate TOC is broken on https://popmotion.io/api/calc/ 

like this

It maybe occur use inline marking.

But sadly, I can't build document on local machine and couldn't check.
If this solution is invalid, please close this.


InventingWithMonster:
Excellent thanks

PR #560, opened 2018-10-13
Fixed config caching for dynamically created configs

### LINKED ISSUE #550: [react-pose] Posed element config is hoisted
Hello,

I wanted to take advantage of the fact that we can initialize a posed element using the props : 

```jsx
const Animation = posed.div(({ from, enter, leave }) => ({
  from,
  enter,
  leave
}))
```

But now everytime I use my `Animation` element (multiple `Animation` per screen), they all have the same animations, even if I pass different props to them.

Here is a log (key + props) of what I pass to each of my `Animation` : 
image

You can see that Layout.FAB have a different animation than the two others, but when I transition everything is animated the same way (it seems to pick the props from the first rendered `Animation`).

I'd expect `Animation` to work like a class, and being instantiated only when I call `<Animation/>`.

**EDIT** : Seems like the `config` is hoisted : https://github.com/Popmotion/popmotion/blob/master/packages/react-pose/src/posed.tsx#L31

Is it necessary? Can we avoid the hoisting when using a function?

Andarist:
fixes #550, cc @jgoux

### DIFF
--- a/packages/react-pose/src/posed.tsx
+++ b/packages/react-pose/src/posed.tsx
@@ -28,20 +28,14 @@ const componentCache = new Map<
 
 const createComponentFactory = (key: string | React.ComponentType) => {
   const componentFactory: ComponentFactory<any> = (poseConfig = {}) => {
-    let config: DomPopmotionConfig;
-
     return ({ withParent = true, ...props }) => {
-      config =
-        config ||
-        (typeof poseConfig === 'function' ? poseConfig(props) : poseConfig);
-
       return !withParent || props.parentValues ? (
-        <PoseElement poseConfig={config} elementType={key} {...props} />
+        <PoseElement poseConfig={poseConfig} elementType={key} {...props} />
       ) : (
         <PoseParentContext.Consumer>
           {(parentCtx: PoseContextProps) => (
             <PoseElement
-              poseConfig={config}
+              poseConfig={poseConfig}
               elementType={key}
               {...props}
               {...parentCtx}
--- a/packages/react-pose/src/components/PoseElement/index.tsx
+++ b/packages/react-pose/src/components/PoseElement/index.tsx
@@ -49,9 +49,34 @@ const objectToMap = (obj: { [key: string]: any }): Map<string, any> =>
 
 const testAlwaysTrue = () => true;
 
+const filterProps = ({
+  elementType,
+  poseConfig,
+  onValueChange,
+  innerRef,
+  _pose,
+  pose,
+  initialPose,
+  poseKey,
+  onPoseComplete,
+  getParentPoseConfig,
+  registerChild,
+  onUnmount,
+  getInitialPoseFromParent,
+  popFromFlow,
+  values,
+  parentValues,
+  onDragStart,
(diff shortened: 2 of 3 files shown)


packages/react-pose/src/posed.tsx
```diff
@@ -26,31 +26,55 @@ const componentCache = new Map<
...
+          </PoseParentContext.Consumer>
+        );
+      }
+    }
+
+    if (process.env.NODE_ENV !== 'production') {
+      (Posed as any).prototype.componentDidUpdate = function (prevProps: PoseElementProps) {
+        warning(
+          shouldRenderWithoutParent(this.props) === shouldRenderWithoutParent(prevProps),
+          "Pose elements once rendered must always render the same way (with or without parent)."
```
Andarist: might not be the best explanation, but didnt know how to rephrase this nicely

InventingWithMonster:
Would it be easier and lighter-weight to make it possible to pass `poseConfig` as a function to `PoseElement` and resolve it in that constructor? Or does that come with an issue that I haven't thought of?

Andarist:
Wouldnt it just move where resolving of the poseConfig function happens?

The difference would be that it would be called with different set of props, not necessarily a bad thing, but in theory might break somebody's code (not very likely, but possible).

Also giving a nice developer warning about changing the withParent prop would become more cumbersome (but not impossible, thanks to WeakMaps 😍 )

InventingWithMonster:
By moving it to `PoseElement` we'd remove the need for caching the config at all because it'd be specific to that class instance. Whereas currently we're caching it like this because it's currently resolved in a render function. It'd be a little lighter too as we'd still be generating a function rather than a class for each `posed.div()`

Wouldn't it still be called with the same props, the props present when the component is initialised?

Andarist:
> By moving it to PoseElement we'd remove the need for caching the config at all because it'd be specific to that class instance. Whereas currently we're caching it like this because it's currently resolved in a render function.

I was referring to the implementation in this PR - which has no config caching as I've converted the component to a class, that's why I've mentioned that it would just move the resolving to the other place.

Comparing to the current code on the master your proposed solution is better than mine in regards to resolving the config 👌

> Wouldn't it still be called with the same props, the props present when the component is initialised?

- master calls the resolver with props minus `withParent`
- current state of this PR calls the resolver with all props
- your solution would call resolver with props plus elementType and poseConfig or with props plus elementType, poseConfig & parentCtx spread over them, both those versions would call it without withParent, so preserving master behaviour, as it would get removed at higher level (we can fight `elementType` & `poseConfig` and `parentCtx` though, we'd just have to remove them with spread in the constructor and call the resolver with that)

Andarist:
I've adjusted the code according to your suggested change.

InventingWithMonster:
Looks great, thanks!

InventingWithMonster:
Published as `react-pose@3.3.7`

jgoux:
Thanks both of you! 🌟

Andarist:
@InventingWithMonster I think you have forgotten to push out the releasing commit, as the master has still react-pose@3.3.6

PR #563, opened 2018-10-14
Fixed issue with applyAtEnd applied with wrong value when transitioni…

Andarist:
…ng from one positional unit type to another


InventingWithMonster:
Ah yep this has gone out in `popmotion-pose@3.3.2`

PR #564, opened 2018-10-14
Handle flipping to zero width & height

### LINKED ISSUE #343: Pose: FLIP to width/height 0
Currently you can't FLIP to width/height 0 because an element with those dimensions can't be inverted with a transform (inverting something with height: 0 via scaleY would result in `scaleY: Infinity`)

We can detect this an instead set width/height to 1px, and then set to `0` on complete

Andarist:
fixes #343

Current downside is that the default animation doesn't look too good - the element stays at 1px height for too long and it's noticeable. Any ideas how to fix it?

### DIFF
--- a/packages/popmotion-pose/src/dom/flip.ts
+++ b/packages/popmotion-pose/src/dom/flip.ts
@@ -28,7 +28,10 @@ const findCenter = ({ top, right, bottom, left }: BoundingBox) => ({
   y: (top + bottom) / 2
 });
 
-const positionalProps = ['width', 'height', 'top', 'left', 'bottom', 'right'];
+const dimensionProps = ['width', 'height'];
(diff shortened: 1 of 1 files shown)


Andarist:
> I wonder if there's a simple way of setting opacity to 0 when the element is at 1?

Transitioning immediately to 0 would have a better effect (I think) - we'd avoid future transition from 1px to 0px that could cause a slight layout jump.

Either way - not sure where and how this should be applied conditionally 😉 

> So the problem is when it animates to 1 it kinda stays there at the end of the animation? Have you tried doing it with a tween to see if the problem is the spring taking too long to "settle"?

That's the most probable cause - linear transition has no weird effect of overstaying its welcome, it just doesnt look that nice 😉

Issue #548, opened 2018-10-09
[discussion] Should onPoseComplete fire for nested elements?


Andarist:
Issue explained in more detail - https://github.com/Popmotion/popmotion/issues/544#issuecomment-427795361

Example - https://codesandbox.io/s/xzxk44jrww .

I'm not sure how exactly this could be utilized, but it seems reasonable that people might expect this callback being fired for nested elements.

nkov:
I believe it should. It makes semantically, that any element which is animating should have the callback fired when it's relevant. For me, the use case is like this:

I am using `react-router` to navigate and I have a child component which animates on the mount. From this discussion, I learned that the `onPoseComplete` event fires on the parent, but I need it to fire in this component so that I can change the local state. The new component which loads has an input box, which slides up from the bottom when the component mounts. This all works great. However I need the `<input>` to be disabled until the the animation is complete (to fix a quick with iOS). To accomplish this, I set a local state: `inputDisabled: true` and expect that `onPoseComplete` will fire when the input finishes sliding up, so that I can set `inputDisabled` to false.

Andarist:
Makes sense to me. Would you be willing to look into implementing this?

nkov:
I haven't written TS before but this seems like a good opportunity to get my feet wet. I can give it a try when I have some time, any pointers on where to start looking?

Andarist:
If you remove `initialPose` then no transition happens on mount, so `onPoseComplete` doesn't fire.

PR #569, opened 2018-10-18
React-pose: add onPoseComplete for children

nkov:
With this, animated children who have `onPoseComplete` will be called when a parent finishes it's animation.

Open to discuss this approach.


nkov:
Addresses #548

packages/react-pose/src/_tests/index.test.tsx
```diff
@@ -118,6 +118,43 @@ test('posed: `onPoseComplete` gets called with pose as argument', () => {
...
+          />
+        </Parent>
+      </PoseGroup>
+    )
+    const wrapper = mount(<Group />)
+    wrapper.setProps({ pose: 'enter' })
+  })
+});
+
+test('posed: `onPoseComplete` gets called for mounting children', () => {
```
Andarist: would be good to add tests for regular posed elements, not only for PoseGroup children

i.e. for such tree
```
<Sidebar initialPose="closed" pose="opened">
  <Content onPoseComplete={() => {}}>
  </Content>
</Sidebar>
```

packages/popmotion-pose/src/dom/pose.ts
```diff
@@ -52,6 +53,7 @@ const createPoseConfig = (
       onDragEnd,
       onPressStart,
       onPressEnd,
+      onPoseComplete,
```
Andarist: is this fixing some wrongly passed prop?

packages/react-pose/src/components/PoseElement/index.tsx
```diff
@@ -273,18 +277,29 @@ class PoseElement extends React.PureComponent<PoseElementInternalProps> {
   }
 
   setPose(pose: CurrentPose) {
-    const { onPoseComplete } = this.props;
     const poseList: string[] = Array.isArray(pose) ? pose : [pose];
 
     Promise.all(poseList.map(key => key && this.poser.set(key))).then(
-      () => onPoseComplete && onPoseComplete(pose)
+      () => this.callCompleteCallbacks(pose)
```
Andarist: I think this might get called at wrong times and for elements which don't use certain poses, u can have a parent which transitions from closed to opened with a child that has no config for `opened` pose and thus I believe the child's onPoseComplete shouldnt get called

packages/popmotion-pose/src/dom/pose.ts
Andarist: oh, you are passing it to the config, currently `onPoseComplete` is not part of the config - it's a prop so it can easily change over time and with your current implementation you'd call stale handlers

Andarist:
It seems to me that you'd have to colocate calling `onPoseComplete` with `poser.set` handlers for each element, so for starters here - https://github.com/Popmotion/popmotion/blob/879e6f0300933947fd6de8ef24403697460a2934/packages/pose-core/src/factories/setter.ts#L75

PR #570, opened 2018-10-19
Allow config as function in react-pose-core

Andarist:
cc @MrLoh


InventingWithMonster:
Excellent man thanks for thjs

InventingWithMonster:
`react-pose-core@0.5.0`
`react-native-pose@0.9.0`

PR #571, opened 2018-10-19
Fixed issue with applyAtEnd being mutated

Andarist:


packages/popmotion-pose/src/dom/unit-conversion.ts
```diff
@@ -47,29 +47,36 @@ const isPositionalKey = (v: string) => positionalValues[v] !== undefined;
...
-export const convertPositionalUnits = (state: PoserState, pose: Pose) => {
+export const convertPositionalUnits = (state: PoserState, nextPose: Pose) => {
   const { values, props } = state;
   const { element, elementStyler } = props;
-  const positionalPoseKeys = Object.keys(pose).filter(isPositionalKey);
+  const positionalPoseKeys = Object.keys(nextPose).filter(isPositionalKey);
   const changedPositionalKeys: string[] = [];
   const elementComputedStyle = getComputedStyle(element);
 
+  let applyAtEndHasBeenCopied = false;
```
Andarist: not the best name, but couldnt figure out a better one quickly :<

packages/popmotion-pose/src/dom/unit-conversion.ts
```diff
@@ -47,29 +47,36 @@ const isPositionalKey = (v: string) => positionalValues[v] !== undefined;
 export const isPositional = (pose: Pose) =>
   Object.keys(pose).some(isPositionalKey);
 
-export const convertPositionalUnits = (state: PoserState, pose: Pose) => {
+export const convertPositionalUnits = (state: PoserState, nextPose: Pose) => {
```
Andarist: unified argument name with `flipPose`

packages/popmotion-pose/src/dom/unit-conversion.ts
```diff
@@ -78,7 +85,6 @@ export const convertPositionalUnits = (state: PoserState, pose: Pose) => {
   elementStyler.render();
 
   const targetBbox = element.getBoundingClientRect();
-  const newPose = { ...pose };
```
Andarist: `pose` was already a copy, so there is no need to copy it again, https://github.com/Popmotion/popmotion/pull/571/files#diff-6575bee1d0a5453bf870f9c62efe8823R83

packages/popmotion-pose/src/dom/flip.ts
```diff
@@ -111,40 +111,34 @@ const implicitlyFlipPose = (state: PoserState, nextPose: Pose) => {
   // Set transform origins
   elementStyler.set({ originX, originY });
 
-  // Create target values
-  const flipPoseProps: FlipPose = {};
```
Andarist: `nextPose` is already a copy, so it should be safe to just mutate it here https://github.com/Popmotion/popmotion/pull/571/files#diff-6575bee1d0a5453bf870f9c62efe8823R83

PR #578, opened 2018-10-24
Minor tweak of PoseParentContext

Andarist:


packages/react-pose/src/components/PoseElement/index.tsx
```diff
@@ -17,7 +17,11 @@ import { invariant } from 'hey-listen';
 import { hasChanged } from '../../utils/has-changed';
 import { pickAssign } from '../../utils/pick-assign';
 
-export const PoseParentContext = createContext({});
+const {
+  Consumer: PoseParentConsumer,
+  Provider: PoseParentProvider
+} = createContext();
+export const PoseParentConsumer;
```
Andarist: I thought it would be better to keep the provider local/private here.

Additionally there is no need to pass a default value to `createContext` as its private and thus always provided here.

packages/react-pose/src/components/PoseElement/index.tsx
```diff
@@ -17,7 +17,11 @@ import { invariant } from 'hey-listen';
 import { hasChanged } from '../../utils/has-changed';
 import { pickAssign } from '../../utils/pick-assign';
 
-export const PoseParentContext = createContext({});
+const {
+  Consumer: PoseParentConsumer,
+  Provider: PoseParentProvider
+} = createContext({});
+export { PoseParentConsumer };
```
Andarist: I thought it would be better to keep the provider local/private here.

InventingWithMonster: Yeah fair

Issue #576, opened 2018-10-24
Replace componentWillUpdate with componentDidUpdate


codepunkt:
`componentWillUpdate` is deprecated since react v*16.3.0*. Can we replace it with `componentDidUpdate`?

Wrapping an app that uses `react-pose` in react's `StrictMode` component notifies about this:

```
index.js:1452 Warning: Unsafe lifecycle methods were found within a strict-mode tree:
    in StrictMode (at App.tsx:41)
    in App (at src/index.tsx:8)

componentWillUpdate: Please update the following components to use componentDidUpdate instead:
PoseElement
```

PR #579, opened 2018-10-24
Use getSnapshotBeforeUpdate instead of UNSAFE_componentWillUpdate hook

Andarist:
Should fix https://github.com/Popmotion/popmotion/issues/576


InventingWithMonster:
👍

PR #580, opened 2018-10-24
Fix forwardRef support

### LINKED ISSUE #512: Fix `isForwardRef` typings
https://github.com/Popmotion/popmotion/pull/509#discussion-diff-219638818R176

Andarist:
fixes #512


packages/react-pose/rollup.config.js
```diff
@@ -12,19 +13,8 @@ const noDeclarationConfig = Object.assign({}, typescriptConfig, {
...
-    'node_modules/react/index.js': [
-      'createContext',
-      'createElement',
-      'Fragment',
-      'Component',
-      'PureComponent',
-      'Children',
-      'cloneElement'
-    ]
+    'react-is': Object.keys(ReactIs),
```
InventingWithMonster: Clever

InventingWithMonster:
@Andarist I wonder if we ditch `innerRef` and `hostRef` in the near future. It only causes upset.

Andarist:
@InventingWithMonster not sure if we can ditch it, we'd have to force using `React.forwardRef` on consumers as in
```js
const Component = React.forwardRef((props, ref) => <Comp {...props} innerRef={ref} />)
const PosedComponent = posed(Component)({})
```

Although maybe that's the way to go 🤔 one can always use `forwardRef`, so technically it might not be a problem - arguably doesnt even cause much more boilerplate than current
```js
const Component = ({ innerRef, ...props }) => <Comp {...props} hostRef={innerRef} />)
const PosedComponent = posed(Component)({})
```

InventingWithMonster:
I think that’s what I like about it - it’s just one consistent way to
explain it. Depends if Styled Components and Emotion support it

Andarist:
> I think that’s what I like about it - it’s just one consistent way to explain it. Depends if Styled Components and Emotion support it

They actually don't suffer from this problem. The only ref-related thing in them was about how to obtain a ref to styled component - so which prop should it accept (innerRef or ref as in `<StyledButton innerRef={myRef} />`). 

Popmotion also has this mentioned problem but we can migrate to `React.forwardRef` internally and recommend using `ref`.

Popmotion is different from styled-components & emotion that it ALSO has to obtain a ref from the wrapped component to "connect" to it, so it's not a simple pass through - it also has to use refs internally even if consumer doesnt pass ref prop to the posed element.

Basically what Im trying to say - styled-components & emotion were only concerned with how to give consumers ref to the DOM elements and Popmotion is concerned with that and with how to obtain a ref for itself too.

Anyway styled-components@4 is already using React.forwardRef internally and upcoming emotion@10 will use it too, that's however not explicitly a concern for consumers (implicitly it is because they need to use different prop to get ref).

My snippets above are showing how Popmotion could get ref in the future for its internal usage which is not tightly coupled to the problem of how to get a ref from posed element.

That being said (I feel i have kinda repeated same thing multiple times here 😅) after giving this a little bit of thought I'm also in favour of just recommending wrapping composite components in `React.forwardRef` if one wants to connect such to Popmotion which would mean that we could just get rid of handling refs in any special way and just use `ref={this.getRef}` in render.

InventingWithMonster:
Yeah that’s a good distinction, although I was just thinking about whether
they use forwardRef or not.

I reckon I’m probably not gonna bother migrating or dropping the existing
method just yet but maybe with a hooks rewrite I’ll ditch the
innerRef/hostRef approach

PR #582, opened 2018-10-24
Migrate from enzyme to react-testing-library

Andarist:
Dont know how you like it, feel free to just close it. I think `react-testing-library` focuses more on public APIs and on how people interact with & use components which is better than testing with implementation-specific enzyme methods.


InventingWithMonster:
Yeah I like it, it's a lot less messy than the enzyme approach.

PR #583, opened 2018-10-24
Build tweaks

Andarist:
- added https://github.com/4Catalyzer/babel-plugin-dev-expression to strip invariant & warning strings in production
- moved annotate-pure-calls & dev-expression to rollup (by adding rollup-plugin-babel)
- added tslib as dependency where it was missing
- fixed some build warning


InventingWithMonster:
Sorry stupid question maybe but have you checked the output to verify this works as intended?

Andarist:
I've only checked that it builds & pass tests - not really that it runs OK in the browser.

Andarist:
@InventingWithMonster I've rebased this and rechecked everything, should be ready to be merged

PR #584, opened 2018-10-24
Mutate newState in place instead of spreading

Andarist:


packages/react-pose/src/components/Transition/children.ts
```diff
@@ -167,10 +167,11 @@ const handleTransition = (
   return nextState;
 };
 
-export default (props: Props, state: State) => ({
-  hasInitialized: true,
-  ...handleTransition(props, state)
-});
+export default (props: Props, state: State) => {
+  const newState = handleTransition(props, state);
```
Andarist: `newState` is always a fresh object, no need to copy it

PR #585, opened 2018-10-24
Short circuit component factory creation during initialization

Andarist:


packages/react-pose/src/posed.tsx
```diff
@@ -60,7 +60,7 @@ const posed = ((component: React.ComponentType | string) =>
   getComponentFactory(component)) as Posed;
 
 supportedElements.reduce((acc, key) => {
-  acc[key] = getComponentFactory(key);
+  acc[key] = createComponentFactory(key);
```
Andarist: the cache is empty when this reduce takes place, so there is no point to use getComponentFactory which checks the cache

InventingWithMonster:
Thanks!

PR #587, opened 2018-10-25
Use Popcorn in the codebase

Andarist:


packages/style-value-types/src/utils.ts
```diff
@@ -1,3 +1,4 @@
+// TODO: dedupe with popcorn, atm it causes a dependency cycle though
```
InventingWithMonster: I'll probably be moving this popcorn at some point which should fix this

InventingWithMonster:
Excellent you're braver than me! Thanks a lot for this one

PR #588, opened 2018-10-25
stylefire: fix transform properties not beeing prefixed

SimonSelg:
In stylefire, the `transform` and the `transform-origin` properties where left out in the prefix logic. 

cheers and thanks for the great library :)

### DIFF
--- a/packages/stylefire/src/css/render.ts
+++ b/packages/stylefire/src/css/render.ts
@@ -120,7 +120,7 @@ export default function buildStylePropertyString(
   // If we have transform origin, set
   if (hasTransformOrigin) {
     propertyString += styleRule(
-      TRANSFORM_ORIGIN,
+        prefixer(TRANSFORM_ORIGIN, true),
       `${state.transformOriginX || 0} ${state.transformOriginY ||
         0} ${state.transformOriginZ || 0}`
     );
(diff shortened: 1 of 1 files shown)


InventingWithMonster:
Hi @SimonSelg 

Is there a browser you encountered where this led to `transform-origin` not being applied?

SimonSelg:
Hi @InventingWithMonster, sorry for taking that long to get back to you.

I encountered only the `transform` property not beeing prefixed on iOS 7/8 (needs -webkit- prefix). I didnt check the `transform-origin` beeing prefixed in a browser, but I didnt see a code path where the `transform-origin` property was getting prefixed.

SimonSelg:
Hey @InventingWithMonster, any updates on this? This PR still has the "Awaiting response" label, is this intended?

Cheers

InventingWithMonster:
Hi @SimonSelg, yeah I ask because Safari on iOS 7&8 are used by a total of %0.23 users https://caniuse.com/usage-table so usually I wouldn't write code to explicitly support these users. The prefixer usually runs because for arbitrary CSS rules we don't know whether these are widely supported or not, but we know `transform` and `transform-origin` are.

However I as a point of principle I agree all CSS rule names should be resolved in the same way at least so 👍

InventingWithMonster:
Published as `stylefire@2.2.0`. Thanks for the PR!

LBRDan:
@InventingWithMonster, sorry bothering you
I certainly agree with you about using the same resolution way to calculate the css rule name prefix
I was wondering why not apply the same thought on #689 ?
Thanks 🙏

PR #594, opened 2018-10-27
Don't override child props with parent props when starting child anim…

### LINKED ISSUE #593: react-pose: applyAtStart/applyAtEnd in inner posed element gets applied to outer posed element
## Describe the bug

When an outer element sets pose and an inner element uses `applyAtStart` or `applyAtEnd` the properties get applied to the outer element instead of the inner element.

Doing something like:

```
const Outer = posed.div({
  open: {},
  closed: {}
});

const Inner = posed.div({
  open: {
    applyAtStart: { display: 'block' }
  },
  closed: {
    applyAtEnd: { display: 'none' }
  }
});
```

And then:

```
<Outer pose="closed"><Inner /></Outer>
```

`Outer` will get `display: none` applied instead of `Inner`.

## How to reproduce

https://codesandbox.io/s/2v8821kpxp contains a full example of where the inner element should toggle visibility and the outer element should stay visible.

## Expected behaviour

`applyAtStart`  and `applyAtEnd` get applied to the element for which they are defined.

Andarist:
…ations

fixes #593

I'll write a test for this later, but I'm not sure if current fix is completely OK - was there any reason why those props were transferred here?

This is wrong because it unconditionally overrides child props with parent's and there are things such as `element`, `elementStyler` there that are used later in example in `applyValues` so they get applied on the wrong element.

I suspect that because this overrides child props then resolvers might get called with wrong values.

### DIFF
--- a/packages/pose-core/src/factories/setter.ts
+++ b/packages/pose-core/src/factories/setter.ts
@@ -73,7 +73,6 @@ const startChildAnimations = <V, A, C, P>(
   Array.from(children).forEach((child, i) => {
     animations.push(
       child.set(next, {
-        ...props,
         delay: delay + generateStaggerDuration(i)
       })
     );


InventingWithMonster:
Yeah I believe this was a fuck up. Nice catch thanks.

PR #596, opened 2018-10-30
some `Array.from` polyfills cannot handle a Set, fixes #418

### LINKED ISSUE #418: react-pose: in IE11 child-components seem to ignore parent-pose keys
## Description
In Internet Explorer 11 there seems to be an issue with react-pose: child-components seem to ignore the current pose-key set in a parent.

I've setup a demo repo to reproduce the issue: https://github.com/klaasman/react-pose-ie11-bug

## How to reproduce
1. `git clone https://github.com/klaasman/react-pose-ie11-bug`
2. `cd react-pose-ie11-bug`
3. `yarn install`
4. `yarn start`
5. open [`http://localhost:8000`](http://localhost:8000) in IE11

## Expected behaviour
In IE 11 child components should inherit the pose defined by a parent component.  
See the [screen captures](https://github.com/klaasman/react-pose-ie11-bug#screen-captures) for a visual result.

## Link to example
https://github.com/klaasman/react-pose-ie11-bug

## Device information
 - Internet Explorer 11

klaasman:
See #418.

### DIFF
--- a/packages/pose-core/src/factories/setter.ts
+++ b/packages/pose-core/src/factories/setter.ts
@@ -70,7 +70,7 @@ const startChildAnimations = <V, A, C, P>(
       ? (i: number) => i * stagger
       : (i: number) => maxStaggerDuration - i * stagger;
 
-  Array.from(children).forEach((child, i) => {
+  Array.from(children.values()).forEach((child, i) => {
     animations.push(
       child.set(next, {
         ...props,


packages/pose-core/src/factories/setter.ts
```diff
@@ -70,7 +70,7 @@ const startChildAnimations = <V, A, C, P>(
       ? (i: number) => i * stagger
       : (i: number) => maxStaggerDuration - i * stagger;
 
-  Array.from(children).forEach((child, i) => {
+  Array.from(children.values()).forEach((child, i) => {
```
Andarist: Maybe just skip `Array.from` altogether? `Set` just has a `forEach` method

klaasman: Correct, but the `Set.prototype.forEach` doesn't come with an index parameter which is used to create a staggered delay:

https://github.com/Popmotion/popmotion/blob/11009eeb33d7d7a98bc5a40b068bb65a2fd2727e/packages/pose-core/src/factories/setter.ts#L73-L80

Andarist: Right, havent noticed this by looking quickly at this. This has caught my eye as potential problem - https://github.com/Popmotion/popmotion/issues/597 , but for different reasons, so it certainly shouldnt block this PR.

InventingWithMonster:
Hey @klaasman thanks for the PR. I think the problem with this approach is it's attempting to fix a bug in IE11 but `Set.values` isn't available in IE11 either. Maybe with a note in the Pose FAQs that it requires a `Set.values` polyfill (along with link)

Andarist:
Figuring out how to deal with https://github.com/Popmotion/popmotion/issues/597 might require some refactor that **could** potentially just remove using Set here as it seems that it's just not good enough  as the collection order matters.

Issue #601, opened 2018-11-01
[react-pose]: React.StrictMode causes removeChild to not be called in Transition component


drobannx:
Happens on version 3.3.7 but this bug did not happen on 3.3.0, changing the version back to 3.3.0 in the codepen will show the expected behavior.

Using the latest version, I noticed issues with elements still being rendered that should not be after animation completes. I debugged the issue and noticed that when I turn off React.StrictMode, the issue went away. I dug deeper and wrote a failing test to try and narrow down the issue. I believe the issue is around the use of getDerivedStateFromProps inside of the `Transition` component and also calling a setState in the `removeChild` function. This causes there to be different sources of truth for the `displayedChildren` state value. I also noticed when debugging my failing test that the `removeChild` function never gets called with StrictMode on but it does with it off.

I'd be happy to submit a PR with a fix since I have written a failing test already but I kind of need some direction how to fix the issue with the way the getDerivedStateFromProps is being used.

## How to reproduce
1. Setup a PoseGroup with a conditional pose.div. Opacity animation is easiest.
2. Surround your PoseGroup with a <React.StrictMode>
3. Trigger something to change the condition
4. See the old element is still being rendered even though it should not.

## Expected behaviour
I expect that even with React.StrictMode on, that the animations would behave the same. If it does not, this may cause issues later on with concurrent react.

## Link to CodePen example (or similar)
https://codesandbox.io/s/404xxjwro7

## Device information
Any device.

Andarist:
Interesting! It seems that this might be a bug in React - according to docs StrictMode shouldnt affect behaviour, only trigger additional checks ( https://reactjs.org/docs/strict-mode.html ) to get you async ready.

Gonna dig into this, thanks for reporting.

InventingWithMonster:
Maybe what they’re really saying is it doesn’t affect documented or
recommended behaviour. Because this is essentially a hack.

Andarist:
I've info about it - https://reactjs.org/docs/strict-mode.html#detecting-unexpected-side-effects

It seems that in dev mode React intentionally invokes some lifecycles twice to expose potentially unsafe side effects to the user. This includes `getDerivatedStateFromProps` - which we use to track how children has changed. Not yet sure how to fix it but as for this use case `getDerivatedStateFromProps` is actually recommended ( https://reactjs.org/docs/react-component.html#static-getderivedstatefromprops ) I think it has to be fixable. 

[...]

PR #604, opened 2018-11-01
Fix issue with reusing stale children

Andarist:
Fixes https://github.com/Popmotion/popmotion/issues/601

The issue was that leaving was reusing stale children - because it came from state (immutable), when React has called getDerivatedStateFromProps twice the first run calculated nextChildren + indexedChildren correctly (trying to put them into state), but second run didnt receive those values because state wasnt updated yet. By mutating `indexedChildren` we fix that issue, making it safer to reuse previous calculations.


InventingWithMonster:
👍 Excellent thanks

PR #605, opened 2018-11-01
Removing innerRef and hostRef support

InventingWithMonster:


packages/popmotion-pose/docs/api/react/posed.md
```diff
@@ -196,9 +202,9 @@ Callbacks that fire when dragging starts or ends. **Note:** These props are immu
 
 Callbacks that fire when pressing starts or ends. **Note:** These props are immutable and can't be changed after mounting.
 
-### innerRef
+### ref
 
-`innerRef?: (ref: Element) => void`
+`ref?: (ref: Element) => void`
 
 An optional function that will call with the posed DOM element when it mounts, and `null` when it unmounts.
```
Andarist: Or ref object

packages/react-pose/CHANGELOG.md
```diff
@@ -2,6 +2,12 @@
 
 React Pose adheres to [Semantic Versioning](http://semver.org/).
 
+## [4.0.0] 2018-11-01
+
+### Removed
+
+- Removed support for `hostRef` and `innerRef` properties.
+
```
Andarist: Was 3.3.8 released? Would be good to release bug fixes without making this breaking change, 4.0 might follow immediately but that would give more time to people for migration

packages/react-pose/src/components/PoseElement/index.tsx
```diff
@@ -211,7 +180,7 @@ class PoseElement extends React.PureComponent<PoseElementInternalProps> {
   componentDidMount() {
     invariant(
       typeof this.ref !== 'undefined',
-      `No DOM ref found. If you're converting an existing component via posed(Component), you must ensure you're passing the hostRef prop to your underlying DOM element.`
```
Andarist: This should validate against ref being a dom element, splitting this into 2 warnings would be great

packages/react-pose/CHANGELOG.md
InventingWithMonster: Yeah it's been released

packages/react-pose/src/_tests/index.test.tsx
```diff
@@ -390,25 +384,21 @@ test('PoseGroup: Provides group props to children on mount', () => {
   let x = 0;
   let y = 0;
 
-  const ChildWithGroupProps = posed(({ innerRef, hostRef, text, ...props }) => (
-    <div ref={innerRef} {...props}>{`text: ${text}`}</div>
-  ))({});
+  const ChildWithGroupProps = posed(
```
Andarist: I think this test could not use `forwardRef` at all - for its simplicity (it doesnt matter to it), I've written it this way because of other issue that in the meantime I've fixed - so now it should be possible to simplify it, what should matter for it is just a property resolver in `enter` pose (which has to be added to the config now)

packages/react-pose/src/components/PoseElement/index.tsx
```diff
@@ -167,33 +163,6 @@ class PoseElement extends React.PureComponent<PoseElementInternalProps> {
...
-      refs.ref = this.setRef;
-    } else {
-      refs.innerRef = this.setRef;
-      refs.hostRef = this.setRef;
-    }
-
-    return refs;
-  };
-
   setRef = (ref: Element) => {
```
Andarist: this has wrong type - should be `Element | null`, I also think that this method body can be unwrapped from this if statement in it, it really shouldnt matter, u either have that function called with `ref` on mount and null on unmount

additionally to provide helpful warnings for the users 2 warnings could get implemented:
- 1 here, validating that `setRef` is not called when `this.ref !== null` (or undefined, not sure whats it initial value)
- second one in `componentDidUpdate` - validating that `this.ref` is not null, this shouldnt happen because we cache the connected element, and its not possible to update it or disconnect it (we'll hold a reference to it anyway in poser)

packages/react-pose/src/posed.tsx
```diff
@@ -28,22 +29,34 @@ const componentCache = new Map<
 
 const createComponentFactory = (key: string | React.ComponentType) => {
   const componentFactory: ComponentFactory<any> = (poseConfig = {}) => {
-    return ({ withParent = true, ...props }) => {
+    // TODO: Replace functional context with new class property API
+    return forwardRef(({ withParent = true, ...props }, ref) => {
+      warning(
```
Andarist: if u want to keep compat for v4 u need to pass `ref || props.innerRef` as `innerRef` to `PoseElement`

InventingWithMonster: This should be provided by ...props, overwriting innerRef={ref} if present

Andarist: Shouldnt ref win over innerRef prop though?

InventingWithMonster: Yeah but I can’t imagine a situation in which they’ll both be defined

PR #611, opened 2018-11-06
Removed console.log from react-pose v4

Albert-Gao:
There seems to have more console.log
Don't know if I should remove them or not.

Just removed the ones that bother me for now.

Correct me if I miss something.


InventingWithMonster:
Yep good catch thanks! I'll reimplement a linting step to prevent this in the future.

InventingWithMonster:
This was released as `4.0.1`, there's also now a TSLint process in place which will catch stuff like this in the future.

PR #612, opened 2018-11-06
Remove dilate util (same as mix)

Andarist:
dilate & mix were the same thing and I think it's better to just keep one of them, personally i like dilate name slightly better because it's more descriptive (for me), although it's not entirely obvious what it does when at first (just by looking at the name) but it makes sense once i learn it

OTOH mix corresponds to mixColor, mixArray etc so maybe we should keep mix and remove dilate?


InventingWithMonster:
Yeah this is a good point, though I'd remove `dilate` instead of `mix`.

Andarist:
K. I'm going to change this PR later to do the opposite.

Andarist:
I've changed the PR to remove `dilate` instead of `mix`

InventingWithMonster:
Nice one thanks for this

InventingWithMonster:
Published as `@popmotion/popcorn@0.2.0`

PR #628, opened 2018-11-13
Direction lock and allow scroll until direction determined

### LINKED ISSUE #618: [React-Pose] A x-draggable element intercepts scroll event on touch devices
## Description
A `draggable` element on the `x` axis in react-pose intercepts vertical scroll events on touch devices.

## How to reproduce
1. On a mobile phone, go over [Codesandbox](https://codesandbox.io/s/p2jryz2roq)
2. Try starting a vertical scroll on the body from the red draggable square
3. The body doesn't scroll

## Expected behaviour

Maybe this is the intended behavior, but I was curious if there was a way that draggable elements would forward scroll trigger events to the DOM. I understand this is not as easy as it sounds but I'm in the situation where this would be useful.

## Codesandbox link
https://codesandbox.io/s/p2jryz2roq

## Device information
 - iPhone X
 - iOS 12
 - Safari

InventingWithMonster:
Fixes #618

### DIFF
--- a/packages/popmotion-pose/src/dom/pose.ts
+++ b/packages/popmotion-pose/src/dom/pose.ts
@@ -36,6 +36,8 @@ const createPoseConfig = (
     onPressStart,
     onPressEnd,
     draggable,
+    directionLock,
+    scrollUntilDragDirection,
     hoverable,
     focusable,
     pressable,
@@ -54,6 +56,8 @@ const createPoseConfig = (
       onPressEnd,
       dragBounds,
       draggable,
+      directionLock,
+      scrollUntilDragDirection,
       hoverable,
       focusable,
       pressable,
@@ -71,6 +75,7 @@ const createPoseConfig = (
       ...poseConfig.drag,
       ...drag
     };
+
     poseConfig.dragEnd = { ...poseConfig.dragEnd, ...dragEnd };
   }
 
--- a/packages/popmotion-pose/src/dom/events.ts
+++ b/packages/popmotion-pose/src/dom/events.ts
@@ -8,6 +8,8 @@ import { DomPopmotionConfig, DomPopmotionPoser } from '../types';
  * event delegation
  */
 
+type CheckPreventDefault = (config: DomPopmotionConfig) => boolean;
+
 type UIEventConfig = {
   startEvents: string;
   endEvents: string;
@@ -16,7 +18,7 @@ type UIEventConfig = {
   startCallback?: string;
   endCallback?: string;
   useDocumentToEnd?: boolean;
-  preventDefault?: boolean;
+  preventDefault?: boolean | CheckPreventDefault;
 };
 
 type UIEventApplicator = (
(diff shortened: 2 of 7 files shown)


packages/popmotion-pose/src/dom/pose.ts
```diff
@@ -36,6 +36,8 @@ const createPoseConfig = (
     onPressStart,
     onPressEnd,
     draggable,
+    directionLock,
```
Andarist: would be helpful to add an example how this config option might be used - with description how the behaviour differs when using it or not

if it only makes sense when using in combination with `scrollUntilDragDirection` then it would be good to add a dev warning in case it gets used without it

meebix:
How is this coming along @InventingWithMonster ? This is a _huge_ enhancement that is needed.

bebensiganteng:
Hate to ask, but saw a new update to the https://github.com/Popmotion/popmotion/blob/master/packages/popmotion-pose/src/dom/pose.ts is that related to this issue?

maximelebreton:
hello @InventingWithMonster, any news about your fix?
Im ready to land a website based on popmotion for a client, but without this key feature, I can't... 

popmotion-scroll-bug
_On this video, dragging on the x-axis works, but on the y-axis doesn't. (I use the mouse scroll to move on the y-axis)_

Thanks for your help! (any quick & dirty fix is welcome!)

InventingWithMonster:
Yep sorry this has been a long time coming. I still don't know how I'm going to get this into Pose, this functionality only currently exists in Framer Motion which is the production-ready software underneath the new Framer API.

I'm thinking of making a migration release from Pose to Framer Motion that accepts both APIs. Pose and Framer Motion share much of the same codebase so adding it wouldn't be like adding a whole second library in terms of filesize. The caveats is you'd need to be using the latest version of React, and Framer Motion is still `0.x` software so the API might change. The docs are here https://www.framer.com/api/animation/ - use `motion.yourHTMLComponent` instead of `Frame`. But if these are acceptable the process would be:

```
yarn add framer-motion
```

```jsx
const Sidebar = () => {
  const x = useMotionValue(-20) // Your initial x with the assumption you've positioned the sidebar off screen via CSS
  const opacity = useTransformValue(x, [-200, 0], [1, 0.2])

  return <motion.div drag dragDirectionLock style={{ x, opacity }} />
}
```

InventingWithMonster:
You can also have the Sidebar `drag="x"` and the content be separate with `drag="y"`, both with `dragDirectionLock` set.

maximelebreton:
Thanks @InventingWithMonster but i'm using `vue-pose`, Framer Motion seems not available on `Vue` (?), and the `dragDirectionLock` parameter not working in `vue-pose`... :'(

In an [old project](https://luuuge.fr/), I fixed this with:
```js
pointer({
  x: albumY.get(),
  preventDefault: false  // <-- this little thing
})
```
Is there a way to put this parameter in `vue-pose`? (even if I need to fork and harcode the fix!)

Thank you!

kessiacastro:
Any news about it or a way to fix that for now?

PR #629, opened 2018-11-13
Fixed issue with getSnapshotBeforeUpdate checking prevProps instead o…

### LINKED ISSUE #614: [react-pose] PoseGroup children animated from original position when a child leaves the DOM
## Describe the bug
As of 3.3.8 when a child is removed from a `<PoseGroup>` the remaining children are animated from their original position. ( 4.0.1 tested as well )

## How to reproduce
1. Go to https://codesandbox.io/s/1y967z8zjj
2. Click on a any item in the menu, this will remove the first child in the list
3. Do this a few times
4. Observe
5. Go to package.json and change to 3.3.7
6. Repeat step 3 & 4, observe the nicer animation.

## Expected behaviour
When children are removed from a <PoseGroup> tree the remaining children should animate from their current position to their new position

## Device information
 - Device: Desktop
 - OS: macOS
 - Browser: Firefox, Chrome
 - Version: Latest

### LINKED ISSUE #624: react-pose version 4 bug with FLIP/PoseGroup not present in version 3
There is a bug in version 4 of react-pose that prevents the FLIP animation running until after a second user input.

To show this I have forked one of the Pose examples on codeSandbox

[working version 3 here](https://codesandbox.io/s/40rr3w4n29)

[bugged version 4 here](https://codesandbox.io/s/6mm19rr3)

version 3 works from the first click but version 4 only works after the second click.

Andarist:
…f this.props

fixes #624 fixes #614

my dumb mistake when I've changed the lifecycle hook - `componentWillUpdate` was called with `nextProps` (having prevProps on `this.props`), with `getSnapshotBeforeUpdate` this is reversed - it's called with `prevProps` and having "nextProps" already on this as props

would love to write a test for this, but I have no idea how :s (except of checking the implementation details in test which I'd like to avoid)

### DIFF
--- a/packages/react-pose/src/components/PoseElement/index.tsx
+++ b/packages/react-pose/src/components/PoseElement/index.tsx
@@ -223,7 +223,8 @@ class PoseElement extends PureComponent<PoseElementInternalProps> {
     }
   }
 
-  getSnapshotBeforeUpdate({ pose, _pose }: PoseElementInternalProps): null {
+  getSnapshotBeforeUpdate(): null {
+    const { pose, _pose } = this.props;
     if (hasPose(pose, 'flip') || hasPose(_pose, 'flip')) this.poser.measure();
     return null;
   }


InventingWithMonster:
Ah such a simple mistake man! Thanks for the fix

InventingWithMonster:
Published as `4.0.2`

Issue #592, opened 2018-10-26
Writing inline styles with `element.style.cssText` violates some CSP policies


Andarist:
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/style-src#Violation_cases

The only way to cover using popmotion with those CSP rules is to switch to using either `node.style[property] = value` or `node.style.setProperty(property, value)`

InventingWithMonster:
Do you know if browsers batch renders for instance if someone was to
animate both width and height with this approach would it relayout twice?
It’s been a while since I looked at this

Andarist:
I think it has been fixed by https://github.com/Popmotion/popmotion/pull/659

PR #634, opened 2018-11-16
Adding Stylefire -> styles object converter

InventingWithMonster:


Andarist:
When refactoring this it would be great to take a look at this - https://github.com/Popmotion/popmotion/issues/592

InventingWithMonster:
@Andarist this should make implementing that much easier but this is a specific feature for something at work so I haven't had a chance to look into actually adding it

PR #640, opened 2018-11-20
Fallback to getBoundingClientRect in stylefire for non-SVGGraphicsEle…

### LINKED ISSUE #639: Stylefire: cannot handle SVG `<stop />` elements (`element.getBBox is not a function`)
## Describe the bug

Stylefire will throw an error when creating a `<stop />` (using `react-pose`).

https://github.com/Popmotion/popmotion/blob/6a02617956616dd0540684f7a92a84dd4b198da1/packages/stylefire/src/svg/index.ts#L39-L40

The `.getBBox()` appears to be a method of the [`SVGGraphicsElement`](https://developer.mozilla.org/en-US/docs/Web/API/SVGGraphicsElement), but the [`SVGStopElement`](https://developer.mozilla.org/en-US/docs/Web/API/SVGStopElement) doesn't implement this `SVGGraphicsElement`.

## How to reproduce
See https://codesandbox.io/s/1m6pnmlnj

```jsx
const Stop = posed.stop()

function App() {
  return (
    <svg>
      <Stop />
    </svg>
  )
}

Andarist:
…ment svg elements

fixes #639 

`getBBox` & `getBoundingClientRect` are not 100% equivalent, but as former is not available for `SVGStopElement` it has to act as "good enough" fallback.

### DIFF
--- a/packages/stylefire/src/svg/index.ts
+++ b/packages/stylefire/src/svg/index.ts
@@ -7,14 +7,14 @@ import getValueType from './value-types';
 
 type SVGProps = {
   dimensions: {
-    x: number,
-    y: number,
-    width: number,
-    height: number
-  },
-  element: SVGElement,
-  isPath: boolean,
-  pathLength?: number
+    x: number;
+    y: number;
+    width: number;
+    height: number;
+  };
+  element: SVGElement;
+  isPath: boolean;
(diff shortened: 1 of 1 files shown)


InventingWithMonster:
Looks good ! Thanks

PR #647, opened 2018-11-26
fix(pose): nested transition fns

### LINKED ISSUE #646: [react-pose] Transition that returns a fn w/ a tween fn inside is not called
## Bug
If you have a transition that returns a callback and in that callback you return a object map which has a callback inside it, its not evaluated. If I return the object directly it works as expected.

### Investigation
If I look at the code, it doesn't look like it evaluates the interior functions if the result was a callback.

```js
var resolveTransition = function (transition, key, value, props, convertTransitionDefinition, getInstantTransition) {
    var resolvedTransition;
    if (typeof transition === 'function') {
        resolvedTransition = transition(props);
    } else if (transition[key] || transition.default) {
        var keyTransition = transition[key] || transition.default;
        if (typeof keyTransition === 'function') {
            resolvedTransition = keyTransition(props);
        } else {
            resolvedTransition = keyTransition;
        }
    } else {
        resolvedTransition = transition;
    }
    return resolvedTransition === false ? getInstantTransition(value, props) : convertTransitionDefinition(value, resolvedTransition, props);
};
```

I think something like this might work - 

```js
  if (typeof transition === 'function') {
    resolvedTransition = transition(props);

    if (typeof resolvedTransition === 'object') {
      resolvedTransition = resolveTransition(resolvedTransition, key, value, props, convertTransitionDefinition, getInstantTransition)
    }
```

### Example

```js
import posed from 'react-pose';
import { tween } from 'popmotion';
import { interpolate } from 'd3-interpolate';

const pathTransition = ({ from, to }) =>
  tween({
    from: 0,
    to: 1
  }).pipe(interpolate(from, to));

export const transition = {
  d: pathTransition,
  default: {
    type: 'spring',
    velocity: 10,
    damping: 100
  }
};

export const PosedLine = posed.path({
  enter: {
    d: ({ enterProps }) => enterProps.d,
    delay: ({ animated, index }) => (animated ? index * 10 : 0),
    transition: (obj) => obj.animated ? transition : { duration: 0 })
  },
```

amcdnl:
Fixes #646

### DIFF
--- a/packages/pose-core/src/factories/setter.ts
+++ b/packages/pose-core/src/factories/setter.ts
@@ -97,6 +97,19 @@ const resolveTransition = <V, A>(
   if (typeof transition === 'function') {
     resolvedTransition = transition(props);
 
+    /**
+     * transition: () => { d: () => tweenFn }
+     */
+    if (typeof resolvedTransition === 'object') {
+      resolvedTransition = resolveTransition(
+        resolvedTransition as TransitionMap<A>,
+        key,
+        value,
+        props,
+        convertTransitionDefinition,
+        getInstantTransition);
+    }
+
     // Or if it's a keyed object
   } else if (transition[key] || transition.default) {
     const keyTransition = transition[key] || transition.default;
--- a/packages/pose-core/src/factories/setter.ts
+++ b/packages/pose-core/src/factories/setter.ts
@@ -100,16 +100,14 @@ const resolveTransition = <V, A>(
     /**
      * transition: () => { d: () => tweenFn }
      */
-    if (typeof resolvedTransition === 'object') {
-      resolvedTransition = resolveTransition(
-        resolvedTransition as TransitionMap<A>,
-        key,
-        value,
-        props,
-        convertTransitionDefinition,
-        getInstantTransition);
-    }
-
+    resolvedTransition = resolveTransition(
+      resolvedTransition as TransitionMap<A>,
+      key,
+      value,
+      props,
+      convertTransitionDefinition,
+      getInstantTransition
+    );
     // Or if it's a keyed object
   } else if (transition[key] || transition.default) {
     const keyTransition = transition[key] || transition.default;
@@ -296,7 +294,8 @@ const createPoseSetter = <V, A, C, P>(
               );
 
               // Add delay if defined on pose
-              const poseDelay = delay || resolveProp(nextPose.delay, transitionProps);
+              const poseDelay =
+                delay || resolveProp(nextPose.delay, transitionProps);
               if (poseDelay) {
                 transition = addActionDelay(poseDelay, transition);
               }


packages/pose-core/src/factories/setter.ts
```diff
@@ -97,6 +97,19 @@ const resolveTransition = <V, A>(
   if (typeof transition === 'function') {
     resolvedTransition = transition(props);
 
+    /**
+     * transition: () => { d: () => tweenFn }
+     */
+    if (typeof resolvedTransition === 'object') {
```
Andarist: according to types I think it cant be non-object, so checking it here is a little bit pointless

Andarist: also when we are talking about types - this change has to be reflected in the typings

Andarist:
Could you add a test for this too?

InventingWithMonster:
Yep happy to merge with some tests

amcdnl:
@Andarist - I removed the check. 2 questions -

1) It looks like the types do work since the `TransitionFactory` is loosely typed to return `any`.

```ts
export type Props = { [key: string]: any };

export type TransitionFactory<A> = (
  props: Props
) => TransitionDefinition | A | false;
```

2) I tried to write a test but its not working correctly. Maybe you can lend me a pointer. 

I added a new function to `testPoser` called `nestedFunctionalDefTransition`. It looks like:

```
  nestedFunctionalDefTransition: {
    x: 5,
    transition: () => ({
      x: () => 10
    })
  },
```
then in the `'resolves custom transitions correctly'` test I added a case after the `functionalDefTransition` that returned the same result but called my new pose instead like:

```
.then(() => {
      expect(testPoser.get().x).toBe(8 * 2)
      return testPoser.set('nestedFunctionalDefTransition')
    })
```

it fails the same with and without my fix to:

```
    expect(received).toBe(expected) // Object.is equality

    Expected: 16
    Received: 50

      244 |     })
      245 |     .then(() => {
    > 246 |       expect(testPoser.get().x).toBe(8 * 2)
          |                                 ^
      247 |       return testPoser.set('nestedFunctionalDefTransition')
      248 |     })
      249 |     .then(() => {

      at _tests/index.test.ts:246:33
```

Any ideas what I'm doing wrong?

Andarist:
1. Im not a TS expert, so im not entirely sure about what im going to say - but maybe we should change the types here, seems like `TransitionDefinition` is too loose, but probably we shouldnt touch it right now. It seems to me that this type is used too often in our definitions - i think it was supposed to be a type of a single transition property but its also used as return type of `TransitionFactory` which is semantically incorrect. I would change this and instead of this:
```ts
  transition?: TransitionMap<A> | TransitionFactory<A>;
``` 
do this 
```ts
type TransitionMapFactory<A> = (props: Props) => TransitionMap<A>;
// ...
  transition?: TransitionMap<A> | TransitionMapFactory<A>;
```
This would still not be ideal, but at least would be slightly more correct and it would make it easier to narrow those types in the future.

2. not sure, could u push out or changes? I could take a look at it.

amcdnl:
@Andarist - Changes are pushed.

@InventingWithMonster - Any thoughts on the types?

Andarist:
@amcdnl could u also push your failing tests? At the moment this PR changes only the implementation file.

Andarist:
This PR can be closed as those commits got merged in as part of the https://github.com/Popmotion/popmotion/pull/714

PR #649, opened 2018-11-27
Adding clamp and ease opts to interpolate

### LINKED ISSUE #630: Popmotion Pure: Interpolate not clamping if output exceeds it's max outputRange despite the docs
## Describe the bug
As stated in https://popmotion.io/api/transformers/#transformers-preset-transformers-interpolate
> Provided values outside the given ranges will be clamped to the output range limits.

## How to reproduce
const invert = interpolate([0, 100], [100, 0]);
invert(155) // -55

## Expected behaviour
invert(155) // 0

## Device information

InventingWithMonster:
Fixes #630

### DIFF
--- a/packages/popcorn/docs/api/utilities/interpolate.md
+++ b/packages/popcorn/docs/api/utilities/interpolate.md
@@ -77,8 +77,50 @@ const mapper = interpolate(
 mapper(0.5); // '15px, rgba(128, 128, 128, 0.5)'
 ```
 
+## Options
+
+`interpolate` takes an optional third argument, options. This is an object that may contain the following configurations:
+
+### Clamp
+
+Clamp values to within the provided ranges.
+
+Default: `true`.
+
+```javascript
+interpolate([0, 1], [0, 100])(2) // 100
+interpolate([0, 1], [0, 100], { clamp: false })(2) // 200
+```
+
+### Ease
+
+Default: `linear`
+
+An easing function, or list of easing functions, to apply to every transition.
+
+If defined as an array, it must be the `input.length - 1`, as it applies to the transitions **between** the defined values.
+
+```javascript
+interpolate(
+  [0, 100, 200],
+  ['#fff', '#500', '#000'],
+  { ease: [linear, easeOut] }
+)
+```
+
 ## Types
 
 ```typescript
-interpolate(input: number[], output: string[] | number[]): string | number;
+interpolate(
+  input: number[],
+  output: string[] | number[],
+  options: InterpolateOptions
+): string | number;
+
+type InterpolateOptions = {
+  clamp?: boolean = true,
+  ease?: Easing | Easing[]
+};
+
+type Easing = v => v;
 ```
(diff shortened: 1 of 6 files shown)


packages/popcorn/docs/api/utilities/interpolate.md
```diff
@@ -77,8 +77,50 @@ const mapper = interpolate(
 mapper(0.5); // '15px, rgba(128, 128, 128, 0.5)'
 ```
 
+## Options
+
+`interpolate` takes an optional third argument, options. This is an object that may contain the following configurations:
+
+### Clamp
+
+Clamp values to within the provided ranges.
```
Andarist: does this (clamping) have to be configurable? what is the use case for turning it off? Cant turning it off just be implemented in userland with `Math.max(nonClampedMax, interpolate(/**/)(/**/))`?

packages/popcorn/src/utils/interpolate.ts
```diff
@@ -18,78 +21,97 @@ const getMixer = (v: number | string) =>
...
+  const inputLength = input.length;
+
+  invariant(
+    inputLength === output.length,
+    'Both input and output ranges must be the same length'
+  );
+
+  // If input runs highest -> lowest, reverse both arrays
+  if (input[0] > input[inputLength - 1]) {
+    input.reverse();
```
Andarist: this mutates the input array in place, which might cause a hard to find bug in users code, would be good to clone those arrays first, before reversing them

packages/popcorn/src/utils/interpolate.ts
```diff
@@ -18,78 +21,97 @@ const getMixer = (v: number | string) =>
...
-        input[0],
-        input[1],
-        output[0] as number,
-        output[1] as number
+  const inputLength = input.length;
+
+  invariant(
+    inputLength === output.length,
+    'Both input and output ranges must be the same length'
+  );
```
Andarist:
```
invariant(
  !ease || !Array.isArray(ease) || ease.length === input.length - 1,
  'Given array of easing functions has to have `input.length - 1` length, as it applies to the transitions **between** the defined values.'
)
```

packages/popcorn/docs/api/utilities/interpolate.md
InventingWithMonster: The use case for turning it off is mapping between two unbounded ranges.

It could probably be achieved in userland but I think for usability most people would prefer a simple config. IMO clamping an unclamped `interpolate` is simpler and more intuitive but I also think `interpolate` will probably be used *mostly* unclamped

Andarist: > MO clamping an unclamped interpolate is simpler and more intuitive but I also think interpolate will probably be used mostly unclamped

If you think it will be used mostly unclamped then making the default of clamp true seems like a worse choice 🤔 

Im worried how this interacts with `ease` here, is the last (or first) easing used for the out of range values then? It makes some sense but according to the docs the easing function describes the easing function to be used BETWEEN sinbling input values.

Andarist: not sure if you have seen this ^, so friendly 🏓 cc @InventingWithMonster

PR #653, opened 2018-11-28
Fixed cancelSync example in docs

dschu-lab:
I just realized that the example from the docs lives at a different place, so I've updated that too!


InventingWithMonster:
Thanks!

PR #659, opened 2018-12-04
Moving Stylefire to Object.assign

InventingWithMonster:
In tests `Object.assign` is multiple times faster than `cssText`

### DIFF
--- a/packages/stylefire/src/css/build-styles.ts
+++ b/packages/stylefire/src/css/build-styles.ts
@@ -113,47 +113,29 @@ const buildStyleProperty = (
   return styles;
 };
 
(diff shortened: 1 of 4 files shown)


packages/stylefire/src/css/build-styles.ts
```diff
@@ -5,7 +5,6 @@ import {
   isTransformProp,
   isTransformOriginProp
 } from './transform-props';
-import prefixer from './prefixer';
```
Andarist: is prefixing happening at some other place now?

InventingWithMonster: Good catch, yeah I did add it elsewhere in `build-styles` but must have lost the change at some point

Issue #541, opened 2018-10-05
Children button "onClick" not triggering when parent div "draggable" = true


yuensunn:
Tried reading the documentation but could not find any solution. Am I missing something?
Currrently using:-
- "react" : "^16.5.2"
- "react-pose": "^3.3.6",

yuensunn:
I just realized that it is working fine on pc but it wont work on mobile.

JavierFVasquez:
Hey, someone has already fixed this bug? I have waiting one month but apearly this bug doesn't matter to anyone

Andarist:
Hm, im unfortunately not sure how to fix it either :s Could u maybe track `touchstart` instead of click and emulate the event yourself? 

This is tricky because we have to call preventDefault but at the moment touchstart happens we cant know if its intended as dragstart or just a click

dbismut:
Hi @andarist, thanks for your answer. I agree, no way to know if it’s a drag or a click on touch start but popmotion preventsDefault on the start and end events for draggable elements no matter what. So in a way it’s already too late to do any check after the element has been instantiated I believe... otherwise yes, we could calculate the drag distance on drag end and prevent default if the distance is different than zero.

Andarist:
We cant determine if we should call preventDefault on touchstart though. Would omitting preventDefault just on touched allow onClick to be called?

dbismut:
Ok so further clarification after some tests: to register a click on iOS, you need to remove `preventDefault` on `touchstart` **and** `touchend`. However, if you do this, document will then register scroll events and therefore when you drag the draggable element you will also scroll the viewport. 

I thought popmotion `multitouch/index.ts` was preventing default on `touchmove` events, but it doesn't seem to have any effect and the viewport keeps scrolling on drag.

I'm not exactly sure why this works, but if you set `preventDefault` to false in the draggable configuration of `popmotion-pose` in `events.ts` and then implement the draggable element as follows:

```js
class Example extends Component {
  onClick = e => {
    console.log('onClick', e);
  };

  componentDidMount() {
    this.el.addEventListener('touchmove', () => {});
   // this should do nothing but it actually prevents the scroll for some reason
  }

  render() {
    return (
      <Box ref={el => (this.el = el)} className="box" onClick={this.onClick}>
        <button onClick={() => alert('click')}>Click me</button>
      </Box>
    );
  }
}
```

... then the draggable registers clicks and doesn't scroll the viewport (on my iPhone at least).

[...]

PR #669, opened 2018-12-13
Allows click event on draggable element on mobile

dbismut:
This should fix https://github.com/Popmotion/popmotion/issues/541#issuecomment-445261912

I'm not exactly sure what's the best practice to add or remove listeners in popmotion, so code review would be much appreciated!

### DIFF
--- a/packages/popmotion-pose/src/dom/events.ts
+++ b/packages/popmotion-pose/src/dom/events.ts
@@ -16,7 +16,7 @@ type UIEventConfig = {
   startCallback?: string;
   endCallback?: string;
   useDocumentToEnd?: boolean;
-  preventDefault?: boolean;
+  preventScroll?: boolean;
 };
 
 type UIEventApplicator = (
@@ -34,7 +34,7 @@ const makeUIEventApplicator = ({
   startCallback,
   endCallback,
   useDocumentToEnd,
-  preventDefault
+  preventScroll
 }: UIEventConfig): UIEventApplicator => (
   element,
   activeActions,
@@ -43,9 +43,18 @@ const makeUIEventApplicator = ({
 ) => {
   const startListener = startPose + 'Start';
   const endListener = startPose + 'End';
+  const moveListener = startPose + 'Move';
+
   const eventStartListener = listen(element, startEvents).start(
     (startEvent: MouseEvent | TouchEvent) => {
-      if (preventDefault) startEvent.preventDefault();
+      if (preventScroll) {
+        const touchMoveListener = listen(element, 'touchmove', {
+          passive: false
+        }).start((e: TouchEvent) => {
+          e.preventDefault();
+        });
+        activeActions.set(moveListener, touchMoveListener);
+      }
 
       poser.set(startPose);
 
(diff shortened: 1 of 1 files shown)


dbismut:
Oh and I'm not exactly sure why CI fails, since it involves `popcorn/utils/interpolate.ts` which is untouched.

Andarist:
Nice! I'm afraid I'm not competent enough to review this properly, so it will have to wait for @InventingWithMonster review.

As to tests - I believe I have fixed this in https://github.com/Popmotion/popmotion/pull/657 . You need to rebase your branch.

Andarist:
@dbismut you've merged upstream branch into yours, which is fine! it works.

Rebase rewrite the git history by applying your new commits on top of the branch on which you want to rebase, co in a sense it fills a gap between some branch from which you have forked with missing commits, but keeps your commits as the most fresh ones. 

Either way is fine 👍

dbismut:
I have, but not extensively. Only on my iPhone X running iOS 12 + Chrome Inspector. Here is the test code: https://codesandbox.io/s/wypk5zl40l (so obviously it only shows the bug on Codesandbox).

Also my understanding is that the `preventDefault` attribute in `events` was only used by `draggable`,  but I'm not exactly sure of any collaterals this PR may have with the rest of the framework.

Also one last thing: although this PR fixes clicks, it doesn't fix inner scroll in overflowing elements inside the `draggable` element. In #541 I've shown how to make it work (commented lines in the Codesandbox link), it would be technically doable to integrate this in `popmotion` directly but I'm not sure it's necessary (it would require looping on the DOM inside the draggable and get elements computed styles and adding listeners to the overflowing div, a bit ugly I guess unless there is a simpler way. Happy to elaborate if you'd like).

dbismut:
@Andarist I'm so so sorry, but I have absolutely no idea how to do this... Is that `git rebase -i master`?

Andarist:
> @Andarist I'm so so sorry, but I have absolutely no idea how to do this... Is that git rebase -i master?

You dont have to do it now, your merge is OK. I was just sharing some knowledge with you. And yeah `git rebase -i master` is one way to go, although it probably should be something like `upstream/master` because `master` is pointing to your local `master` branch.

InventingWithMonster:
I'm not so concerned about scrollable elements within draggables as I think it's generally a bad idea, whereas this is a clear use case. Otherwise this looks good to me.

dbismut:
@InventingWithMonster I had a situation where I used a `draggable` as a full page on mobile to be able to fully control the way it behaves on drag, release, and have functionalities close to what you can get on iOS native apps. So I needed to make sure overflowing inner-divs remained scrollable horizontally. But I agree, probably not a very common use case.

InventingWithMonster:
Published in `popmotion-pose@3.4.1`

JavierFVasquez:
If i want the upgraded features in 3.4.1 version , but in react-pose, what i can do?

InventingWithMonster:
As a minor it should be enough to reinstall React-pose?

dbismut:
@InventingWithMonster @Andarist Hey, I just realized my PR didn't prevent body scrolling on touch devices anymore when dragging a Drag element. I have a fix here https://github.com/Popmotion/popmotion/pull/706

Reaaaally sorry about this.

Issue #179, opened 2017-05-02
Firefox not working at all with v3.5.0 or v4.3.4?


cameronwh:
Hi guys,

We are using version 3.5.0 on our site, to animate a spinning tick icon. Identical to this one: http://codepen.io/popmotion/pen/gamwav

I have tried updating this codepen to use the latest version offered in your /assets/js directory, which is 4.3.4 however it still seems to not be working. I am unsure of what the issue is, as Firefox is just reporting an NS_ERROR_FAILURE, which I believe is a generic error. It works perfectly in Edge & Chrome, so I am a little lost.

Any ideas?

InventingWithMonster:
Hi @cameronwh 

Sorry for the delayed reply, I've been on holiday the last couple of weeks. The latest version is `6.3.5`, though there have been substantial API changes since `4.x`. The NS_ERROR_FAILURE can occur in Firefox if the code is executed before the SVG element has been successfully appended to the DOM (solved with a 'window.ready' event), if it's a mask, or if it's otherwise invisible:

https://www.google.co.uk/search?q=ns_error_failure+svg+firefox&rlz=1C5CHFA_enGB693GB695&oq=ns_error_failure+svg+firefox&aqs=chrome..69i57.8044j0j4&sourceid=chrome&ie=UTF-8

This has been a long-standing limitation with Firefox and SVGs sadly

cameronwh:
Hi @InventingWithMonster thanks for the reply. 

I didn't know about the NS_ERROR_FAILURE - cheers for the heads up. I'll close this now as it's not related to popmotion.

Issue #174, opened 2017-02-10
SVG Path example won't run on Firefox


Mike-NetCode:
I am trying to open https://codepen.io/popmotion/pen/MKNzRR sample in latest firefox (51) and it simply doesn't show up yet the console is clean. The same issued I observed on FF 48.

I tried to use older versions of codepen (3.x and 4.x) with their respective samples of the same animation (tick in a circle) and I got NS_ERROR_FAILURE in the console log, but I dont get any error in 5.x

Can you confirm?

InventingWithMonster:
Hey Mike - I'll upgrade this example to 6.0 and fix the path issue in Firefox

Mike-NetCode:
Oh, I must have been tired as I thought that 5.0 is the latest :)

InventingWithMonster:
@Mike-NetCode Hey Mike. I've updated the pen to 6.0 and works in Firefox now.

The problem was animating `path` tags within `defs`. They don't technically have a bounding box in Firefox hence the error. To fix, I moved to normal paths.

Mike-NetCode:
Thanks I will check it out in a few hours

PR #670, opened 2018-12-16
Stylefire: Add graceful degradation for NS_ERROR_FAILURE errors in Firefox

niieani:
Trying to create a `Styler` on a not rendered SVG Element under Firefox currently results in a hard failure:

```
[Exception... "Failure"  nsresult: "0x80004005 (NS_ERROR_FAILURE)"]
```

The affected `Element`s may be nested under `mask`s, `clipPaths` or ones with effective `display: none` style.

Since there's no reliable way to determine if an element is currently rendered (or not), I've opted for feature detection. If we detect that we cannot measure a not rendered SVG Element (`getBBox()`), we wrap that call in a try/catch (i.e. only in Firefox, so we do not cause a performance regression in other browsers).

An alternative solution to the try/catch is to check if the SVG Element is a (grand)child of an unsupported Element, like this (only `clipPath` below):

```js
/**
 * in Firefox, when element is inside a <clipPath> Element
 * running element.getBBox() will throw an "NS_ERROR_FAILURE" error
 */
export const isInsideClipPath = (element: SVGElement | SVGPathElement) => {
  const { ownerSVGElement } = element;
  if (!ownerSVGElement) return false;
  let parentElement: SVGElement | HTMLElement | null = element;
  while (parentElement && parentElement !== ownerSVGElement) {
    if (parentElement.tagName === 'clipPath') return true;
    parentElement = (element.parentElement as SVGElement | HTMLElement | null);
  }
  return false;
}
```

But I think this adds complexity and won't work in all cases anyway (detecting `display: none` is very hard). `try/catch` shouldn't be too expensive in modern browsers anyway.

This mostly fixes the issue, however in some cases it might work differently across browsers, since we do not get the actual `x`/`y` values.

I didn't add tests since there's actually no way to meaningfully test this without running Firefox.

Solves: #174, #179.

### DIFF
--- a/packages/stylefire/src/svg/utils.ts
+++ b/packages/stylefire/src/svg/utils.ts
@@ -1,0 +1,36 @@
+import { Dimensions } from './types';
+
+/**
+ * Feature detection for getBBox call on an unrendered SVG Element.
+ * Firefox will throw "NS_ERROR_FAILURE" error.
+ * We want to detect this upfront so we do not crash in any future measurement.
+ */
+export const CAN_MEASURE_NOT_RENDERED_SVG_ELEMENTS = (() => {
+  try {
+    document.createElementNS('http://www.w3.org/2000/svg', 'rect').getBBox();
+  } catch (e) {
+    return false;
+  }
+  return true;
+})();
+
+export const getDimensions = (
+  element: SVGElement | SVGPathElement
+): Dimensions =>
+  typeof (element as SVGGraphicsElement).getBBox === 'function'
+    ? (element as SVGGraphicsElement).getBBox()
+    : (element.getBoundingClientRect() as DOMRect);
+
+export const getSVGElementDimensionsSafely = (
+  element: SVGElement | SVGPathElement
+): Dimensions => {
+  if (CAN_MEASURE_NOT_RENDERED_SVG_ELEMENTS) {
+    return getDimensions(element);
+  }
+  try {
+    return getDimensions(element);
+  } catch (e) {
+    // most likely trying to measure an unrendered element under Firefox
+    return { x: 0, y: 0, width: 0, height: 0 };
+  }
+};
--- a/packages/stylefire/src/svg/utils.ts
+++ b/packages/stylefire/src/svg/utils.ts
@@ -21,7 +21,7 @@ export const getDimensions = (
     ? (element as SVGGraphicsElement).getBBox()
     : (element.getBoundingClientRect() as DOMRect);
 
-export const getSVGElementDimensionsSafely = (
+export const getSVGElementDimensions = (
   element: SVGElement | SVGPathElement
 ): Dimensions => {
   if (CAN_MEASURE_NOT_RENDERED_SVG_ELEMENTS) {
--- a/packages/stylefire/src/svg/utils.ts
+++ b/packages/stylefire/src/svg/utils.ts
@@ -1,19 +1,5 @@
 import { Dimensions } from './types';
 
-/**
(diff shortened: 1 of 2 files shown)


packages/stylefire/src/svg/utils.ts
```diff
@@ -0,0 +1,36 @@
...
+})();
+
+export const getDimensions = (
+  element: SVGElement | SVGPathElement
+): Dimensions =>
+  typeof (element as SVGGraphicsElement).getBBox === 'function'
+    ? (element as SVGGraphicsElement).getBBox()
+    : (element.getBoundingClientRect() as DOMRect);
+
+export const getSVGElementDimensionsSafely = (
```
InventingWithMonster: `getSVGElementDimensions` will do. We should assume it's safe (even though it previously wasn't)

InventingWithMonster:
Hey man thanks for this! Great PR. Just one naming thing and then 👍

packages/stylefire/src/svg/utils.ts
```diff
@@ -0,0 +1,36 @@
...
+  element: SVGElement | SVGPathElement
+): Dimensions => {
+  if (CAN_MEASURE_NOT_RENDERED_SVG_ELEMENTS) {
+    return getDimensions(element);
+  }
+  try {
```
Andarist: is try/catch really that expensive nowadays? cant we just try/catch it always without calculating `CAN_MEASURE_NOT_RENDERED_SVG_ELEMENTS` upfront?

packages/stylefire/src/svg/utils.ts
niieani: Will change! Thanks @InventingWithMonster

packages/stylefire/src/svg/utils.ts
Andarist: According to Jason Miller (developit) its no longer an eager deopt and can be safely used (try/catch)

niieani: @Andarist @InventingWithMonster should I simplify the PR to always use try/catch, or do we go with it as is?

InventingWithMonster: @niieani Yeah simplification is preferred, fewer potential sources of bugs.

niieani: Switched to always try/catch.

niieani:
hey @InventingWithMonster & @Andarist, just wanted to ping you in case you didn't see my update to the PR yet. I'd love to drop my local fork :)

Happy New Year! 🎉

InventingWithMonster:
Looks excellent, thanks for your help

InventingWithMonster:
Published as `stylefire@2.3.5`

niieani:
Thanks @InventingWithMonster!

xavierfoucrier:
Hi @InventingWithMonster!

Can you **add this fix to the popmotion package** by updating stylefire from @2.3.4 to @2.3.5?
May be there are other fix to add to the master branch :thinking: 

Thanks a lot :beers: :yellow_heart:

Andarist:
@xavierfoucrier The dependency is specified with a caret - https://github.com/Popmotion/popmotion/blob/b1e2a655a5d7593d66ee1503044ad4033da4df41/packages/popmotion/package.json#L55 , so you should be able to do a fresh install with the newest stylefire as it was a patch bump for that package.

xavierfoucrier:
Oops.. that's right, I missed that, sorry and thanks for the reply :smile:

PR #683, opened 2019-01-04
Move all children in <Transition />

### LINKED ISSUE #682: React Pose: PoseGroup does not animate elements before the one being removed
## Describe the bug
Not all elements are being animated when a PoseGroup child is removed / added

## How to reproduce
1. Go to https://codesandbox.io/s/m4j1py31xx
2. Click on one of the middle elements
3. You can see elements on the right smoothly translating to their new position while elements on the left jump to their new position without animation

This behavior is even more obvious when clicking on the leftmost or rightmost element

## Expected behaviour
I'd expect all elements to be transitionned to their new position

## Device information
 - OS: macOS
 - Browser: chrome, safari
 - Version 4.0.4

esquevin:
Fixes #682

### DIFF
--- a/packages/react-pose/src/components/Transition/children.ts
+++ b/packages/react-pose/src/components/Transition/children.ts
@@ -95,12 +95,8 @@ const handleTransition = (
 
   const moving = new Set(
     prevKeys.filter((key, i) => {
-      if (entering.has(key)) {
-        return false;
-      }
-
-      const nextIndex = nextKeys.indexOf(key);
-      return nextIndex !== -1 && i !== nextIndex;
+      // if it's not entering or leaving
+      return !entering.has(key) || leaving.indexOf(key) === -1;
     })
   );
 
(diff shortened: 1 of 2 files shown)


packages/react-pose/src/components/Transition/children.ts
```diff
@@ -100,7 +100,7 @@ const handleTransition = (
       }
 
       const nextIndex = nextKeys.indexOf(key);
-      return nextIndex !== -1 && i !== nextIndex;
+      return nextIndex !== -1;
```
InventingWithMonster: Is the resulting code any different to `leaving.indexOf(key) !== -1` basically saying, if it isnt entering or leaving?

Andarist:
Would you be able to write a test for this one? Not sure how hard it is at the moment though.

esquevin:
I've added a test and slightly changed the code to better reflect that `moving = not (entering or leaving)`

Please take another look.

InventingWithMonster:
We might need to optimise this approach for longer lists, but IMO looks good to me

Issue #685, opened 2019-01-05
[React-Pose - <Transition>] Insertion index for leaving children doesn't take entering children into consideration


JabX:
Hello,
I'm a new user of `react-pose` and I'm currently rewriting my existing `react-motion` code with it, and it's truly a pleasure.

While using `<Transition>` to animate entering and leaving children in a list, I noticed that nothing was implemented to handle properly both entering and leaving children during the same render, as seen [here](https://github.com/Popmotion/popmotion/blob/master/packages/react-pose/src/components/Transition/children.ts#L150).

The problem arises when a leaving child is located after an entering child, so a quick fix I thought of that seems to work well (at least for my particular issue) was to shift the inserting index of the leaving child by the number of entering children that are located before the original index.

I have a PR with the change that you can see below, and I'd be happy to contribute so that it can be merged.

InventingWithMonster:
Fixed in 4.0.5

InventingWithMonster:
Yeah not sure how that happened, I thought I only posted once(!)

PR #686, opened 2019-01-05
Improve insertion index in <Transition> for leaving children

JabX:
See #685

### DIFF
--- a/packages/react-pose/src/components/Transition/children.ts
+++ b/packages/react-pose/src/components/Transition/children.ts
@@ -141,14 +141,16 @@ const handleTransition = (
         })
       : child;
 
-    const insertionIndex = prevKeys.indexOf(key);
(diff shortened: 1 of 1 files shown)


InventingWithMonster:
Love the commit message. Looks good at first glance though its been a while since I've used this code. Is there any instances you can think of where this approach might break?

JabX:
I am a bit unsure about how best to handle the edge case where both entering and leaving children are at the same index (using `<=` instead `<` changes the order), but I don't think it matters all that much. In my case neither where acceptable so I "disabled" that case by not swapping the component. 

Otherwise I think it should work fine, the cases that worked well are unchanged and at the very least some new ones now work. I can't see it "breaking" but I won't claim that the entire issue is fixed without actual tests that cover every edge case

PR #688, opened 2019-01-08
Return ComponentType<PoseElementProps & T> from Component factory

### LINKED ISSUE #678: React-Pose: Component factory type issue with props
Hey @InventingWithMonster 
:1st_place_medal: Have to say awesome work on popmotion !!!

In popmotion/packages/react-pose/src/posed.tsx
```
export declare type ComponentFactory<T> = 
(poseConfig?: DomPopmotionConfig | DomPopmotionConfigFactory<T>) => ComponentType<any>;

```
does'nt it makes sense to return `ComponentType<PoseElementProps & T>` instead of `ComponentType<any>`

Is this somthing to do with not setting styles directly but via config? 

Thx
Jaspreet

jas99:
fixes #678

### DIFF
--- a/packages/react-pose/src/posed.tsx
+++ b/packages/react-pose/src/posed.tsx
@@ -14,7 +14,7 @@ type DomPopmotionConfigFactory<T> = (
 
 export type ComponentFactory<T> = (
   poseConfig?: DomPopmotionConfig | DomPopmotionConfigFactory<T>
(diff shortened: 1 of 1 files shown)


InventingWithMonster:
Nice one thanks for this

InventingWithMonster:
Published as `react-pose@4.0.6`

Issue #644, opened 2018-11-23
[react-pose] applyAtStart/applyAtEnd problem


AlexeyBee:
I want my accordion to be `display: none` when it is closed. but the animation breaks when I open the accordion. What am I doing wrong?

I want my animation to be smooth

here is my configuration:
```
const Content = posed.div({
  closed: {
    height: 0,
    display: "none",
    applyAtStart: { display: "block" },
    applyAtEnd: { display: "none" }
  },
  open: {
    height: "auto",
    display: "block",
    applyAtStart: { display: "block" },
    applyAtEnd: { display: "block" }
  }
});
```

CodeSandbox example: https://codesandbox.io/s/20575z7ljy
package: "react-pose": "4.0.2"

InventingWithMonster:
This looks like it works:

```
const Content = posed.div({
  closed: {
    height: 0,
    applyAtEnd: { display: "none" }
  },
  open: {
    height: "auto",
    applyAtStart: { display: "block" },
  }
});
```

AlexeyBee:
> This looks like it works:
> 
> ```
> const Content = posed.div({
>   closed: {
>     height: 0,
>     applyAtEnd: { display: "none" }
>   },
>   open: {
>     height: "auto",
>     applyAtStart: { display: "block" },
>   }
> });
> ```

if you click even more, it will break

Issue #698, opened 2019-01-16
React Pose: applyAtStart / applyAtEnd display: none / block preventing transition


neagle:
## Describe the bug
Using enter/exit poses, and using `applyAtStart: { display: 'block' }` and `applyAtEnd: { display: 'none' }` (which is a highly useful pattern), transforming the `y` value (at least) fails to transition for the exit → enter pose. Enter → exit seems to work as expected.

## How to reproduce
1. Go to [this sandbox](https://codesandbox.io/s/6xr2oylyk3)
2. Observe that the Notification fades in (opacity transition), but does not transition downward.
3. Observe that the Notification fades out _and_ transitions upward on exit.
4. In `index.js`, comment out the `applyAtStart` and `applyAtEnd` keys in the Notification pose config.
5. Note that the initial transition now fades in and transitions downward, as desired.
6. If you change `display: 'none|block'` to `visibility: 'hidden|visible'`, the transition also seems to work properly. This seems uniquely related to an element that was set to `display: none`.

## Link to CodeSandbox example
https://codesandbox.io/s/6xr2oylyk3

## Device information
In which environment are you reproducing the bug?
 - Device: MBP
 - OS: OSX
 - Browser: Chrome
 - Version: 71.0.3578.98

## Additional context
I apologize if I am missing anything obvious. Thanks for this tremendous library!

Andarist:
Seems like this might be a duplicate of https://github.com/Popmotion/popmotion/issues/491 . Could you @neagle confirm / deny it? 

I'll try to think about this to see if I can fix it.

PR #703, opened 2019-01-20
Fix/apply at start

### LINKED ISSUE #491: Pose-core: flip & applyAtStart clash
## Describe the bug
It's impossible to make an entering flip animation for posed elements hidden with `applyAtEnd`, because of those lines:

https://github.com/Popmotion/popmotion/blob/5df3b13e8e7a32e8d502a9f1dff1454e01469a21/packages/pose-core/src/factories/setter.ts#L210-L231

`transformPose` call first calculates the flip by calling `explicitlyFlipPose` and only later `applyAtStart` gets applied. The intuition would tell me that this should work and that `applyAtStart` should play nicely with flip here, OTOH `applyAtEnd` & `applyAtStart` ideally would behave the same and assuming my intuition is correct here for `applyAtStart` then fixing this would create an opposite problem of not being able to hide the element with flip.

So maybe it's just matter of documenting this behaviour?

## How to reproduce
```js
const Box = posed.div({
	visible: {
		width: '20vw',
		height: '40vh',
		flip: true,
		applyAtStart: {
			display: 'block',
		},
	},
	hidden: {
		width: '1px',
		height: '1px',
		flip: true,
		applyAtEnd: {
			display: 'none',
		},
	},
})
```

Andarist:
Fixes #491 & possibly #698 too.

### DIFF
--- a/packages/pose-core/src/factories/setter.ts
+++ b/packages/pose-core/src/factories/setter.ts
@@ -206,17 +206,7 @@ const createPoseSetter = <V, A, C, P>(
     const getParentAnimations = (): AnimationsPromiseList => {
       if (!nextPose) return [];
 
-      if (transformPose) nextPose = transformPose(nextPose, next, state);
-
-      const {
-        preTransition,
-        transition: getTransition,
-        applyAtStart,
-        applyAtEnd
-      } = nextPose;
-
-      // Run pre-transition prep, if set
-      if (preTransition) preTransition(baseTransitionProps);
+      const { applyAtStart } = nextPose;
 
       // Apply initial values, if set
       if (applyAtStart) {
@@ -227,7 +217,15 @@ const createPoseSetter = <V, A, C, P>(
           setValue,
           setValueNative
         );
+        baseTransitionProps.elementStyler.render();
       }
+
+      if (transformPose) nextPose = transformPose(nextPose, next, state);
+
+      const { preTransition, transition: getTransition, applyAtEnd } = nextPose;
+
+      // Run pre-transition prep, if set
+      if (preTransition) preTransition(baseTransitionProps);
 
       const animations = Object.keys(getPoseValues(nextPose)).map(key => {
         // Add pose to the pose order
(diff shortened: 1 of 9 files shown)


packages/pose-core/src/factories/setter.ts
```diff
@@ -227,8 +217,16 @@ const createPoseSetter = <V, A, C, P>(
           setValue,
           setValueNative
         );
+        baseTransitionProps.elementStyler.render();
```
Andarist: this is needed because otherwise those 2:
https://github.com/Popmotion/popmotion/blob/916df7350307d3127bb2dbdd9ea721e03888791c/packages/popmotion-pose/src/dom/unit-conversion.ts#L55
https://github.com/Popmotion/popmotion/blob/916df7350307d3127bb2dbdd9ea721e03888791c/packages/popmotion-pose/src/dom/unit-conversion.ts#L81
would return outdated computed styles.

Most likely the same happens in flip.ts

Andarist: Hm, seems like elementStyler is DOM-specific - so this call has to be moved out of `pose-core`.

How would you suggest doing this?

InventingWithMonster: Ugh. Yeah this makes sense. 

FLIP was a mistake.

You can add "slots" to `pose-core`, that pose flavours can hook into. The same way there's

```
if (transformPose) nextPose = transformPose(nextPose, next, state);
```

You could add an optional function to the `pose-core` factory that's something like "forceRender", kinda generic, that in `popmotion-pose` is configured to be `props.elementStyler.render()` and you call it if you've set `applyAtStart`

Andarist: Cool, was thinking about such hook myself, couldnt settle on a name - `forceRender` sounds nice 👌 

> FLIP was a mistake.

Why? :p do you mean using this technique or its current implementation?

Andarist: Also - the quirk of this solution is that `applyAtStart` cant be transformed (while ie `applyAtEnd` actually is transformed in some cases). It is how it is though, cant figure out any legit use case for transforming `applyAtStart` anyway.

Andarist:
@InventingWithMonster added optional `forceRender` to poseFactory, should be ready to be merged in

Andarist:
This also fixes #644 . That demo still doesnt look too nice though, because of this https://github.com/Popmotion/popmotion/pull/564#issuecomment-430957541

packages/pose-core/src/factories/setter.ts
InventingWithMonster: Mostly because I think 80% of all the bugs we've seen in Pose have been around FLIP!

InventingWithMonster:
Awesome thanks

PR #706, opened 2019-01-22
Fix body scroll

dbismut:
This prevents document.body from scrolling when dragging a Drag element.


gorhgorh:
ahh, now i know why a draggable item constrained to X prevent Y scroll, make sense but prevent some things that could be good to do with pose (like carrousel for example)

would you be interested to get this lock optional rather than mandatory ?

dbismut:
@gorhgorh I agree this would make sense, but the current API doesn't make it easy to pass options when creating a Pose element AFAIK. You can have a look and try though!

PR #708, opened 2019-01-25
Fixed issue with applyAtEnd resetting position of dragged element

### LINKED ISSUE #704: dragBounds does not work properly with v4.0.5
I was trying to recreate the given example in the docs. But It does not work when the dragBounds are set in the latest version **(4.0.5)**. Please check this sandbox

https://codesandbox.io/s/l75vorypm7

Andarist:
fixes #704


packages/popmotion-pose/src/dom/pose.ts
```diff
@@ -16,14 +17,23 @@ import {
 import { isPositional, convertPositionalUnits } from './unit-conversion';
 export { Action, Poser, ValueReaction, ColdSubscription };
 
+const getCurrent = (prop: string) => ({ elementStyler: Styler }) =>
```
Andarist: it's kinda a hack, currently dragEnd results in 2 property sets - one from `just` action and one from `applyAtEnd`, this part of the code tries to trick pose setter to handle it correctly by setting x/y on the poses, so it is what it is

PR #710, opened 2019-01-26
Remove unnecessary guards for addListenerToValue

Andarist:
https://github.com/Popmotion/popmotion/blob/c0936014e36faa77660460042d6e8e384f899203/packages/popmotion-pose/src/types.ts#L70

this is not optional


InventingWithMonster:
Thanks!

PR #714, opened 2019-01-27
Transition map factory fix

Andarist:
This should allow for stricter type checking of used `transition`s. It's built on top of the @amcdnl fix for #646


amcdnl:
Thanks for adding that in !!!!

PR #718, opened 2019-01-29
Fixed react-pose-text build by assigning to instance fields asap and …

Andarist:
…thus relaying on inference (not having to worry about type compatibility)

### DIFF
--- a/packages/react-pose-text/src/SplitText.tsx
+++ b/packages/react-pose-text/src/SplitText.tsx
@@ -1,4 +1,4 @@
-import React, { PureComponent, ReactElement } from 'react';
+import * as React from 'react';
 import posed from 'react-pose';
 import { invariant } from 'hey-listen';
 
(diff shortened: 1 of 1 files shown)


packages/react-pose-text/src/SplitText.tsx
```diff
@@ -37,27 +37,14 @@ const getPoseProps = (props: Props) => {
...
-  Char: (props: PoseProps) => ReactElement<any>;
-
-  constructor(props: Props) {
-    super(props);
-
-    this.state = {};
-
-    const { wordPoses, charPoses, children } = props;
-
-    parseText(children);
```
Andarist: seems like a leftover from the previous implementation, this function is pure and its result was unused here so it was safe to remove it (it was also safe to remove it as it's called inside `getDerivedStateFromProps` which gets called before EACH render, including the initial one, so there is no need to call it in the constructor)

InventingWithMonster:
Thanks!

PR #725, opened 2019-02-04
Fix typo in react hover example.

pleunv:
The react hover example seems to have been based on the press example, this PR replaces a "press" reference with "hover".


InventingWithMonster:
Nice catch, thanks!

PR #730, opened 2019-02-11
Fix merging of default drag/dragEnd configs with the ones specified b…

### LINKED ISSUE #729: React-Pose:  Draggable example no longer works
Per the documentation, the dragEnd should allow you to specify a location to transition to after the drag is released.  Instead, it chooses the location of the release of the drag.

1. Pull from: https://github.com/TonyLB/react-pose-bug
2. npm start

Expect the div in the middle of the page to always try to return to center when released.  It does not.

Andarist:
…y the user

fixes #729

### DIFF
--- a/packages/popmotion-pose/src/dom/pose.ts
+++ b/packages/popmotion-pose/src/dom/pose.ts
@@ -80,11 +80,8 @@ const createPoseConfig = (
   if (draggable) {
     const { drag, dragEnd } = dragPoses(draggable);
 
-    poseConfig.drag = {
-      ...poseConfig.drag,
-      ...drag
-    };
-    poseConfig.dragEnd = { ...poseConfig.dragEnd, ...dragEnd };
+    poseConfig.drag = { ...drag, ...poseConfig.drag };
+    poseConfig.dragEnd = { ...dragEnd, ...poseConfig.dragEnd };
   }
 
   return poseConfig;


InventingWithMonster:
Thanks! Published as `react-pose@4.0.7`

Andarist:
Not quite 😉 The fix was in `popmotion-pose` but you have only released `react-pose`

Florian007:
In react-pose@4.0.8 popmotion-pose needs to be manually upgraded to min version popmotion-pose@3.4.7

designbyadrian:
@Florian007 I've added popmotion-pose as a dependency and reinstalled react-pose, but still have this issue.

Florian007:
> @Florian007 I've added popmotion-pose as a dependency and reinstalled react-pose, but still have this issue.

I do it automatically in the npm postinstall after every install to avoid doing it manually with the npm postinsall hook. You could add this to your scripts section in your package.json like so:

`"postinstall": "cd node_modules/react-pose && npm i popmotion-pose@3.4.7 && cd ../../"`

or do it manually.

designbyadrian:
I tried that, then realised that with Yarn, you can add a resolutions field to package.json that does the same:

```json
dependencies: {  },
resolutions: {
  "popmotion-pose": "3.4.7"
}
```

Sadly, it didn't make any difference. `dragEnd` still doesn't work at all.

PR #744, opened 2019-02-27
Add bounce easings

### LINKED ISSUE #743: Bounce Easing
**Is your feature request related to a problem? Please describe.**
The request is for a bounce type of easing, as described on https://easings.net/ (e.g. easeOutBounce). GreenSock is an example of another library with this functionality. As I use Popmotion for all my animation work, it would be great to have these easings in Popmotion too.

**Describe the solution you'd like**
Either built-in bounce easings, or allowing us to make new easings with Penner functions.

**Describe alternatives you've considered**
I have tried to manually implement Penner functions, but it is messy and not ideal when applying several different animations in parallel. See my CodePen example: https://codepen.io/magnusriga/pen/jJPJzB?editors=0111

Andarist:
closes #743

### DIFF
--- a/packages/easing/src/index.ts
+++ b/packages/easing/src/index.ts
@@ -72,6 +72,34 @@ export const backInOut = mirrored(backIn);
 
 export const anticipate = createAnticipateEasing(DEFAULT_OVERSHOOT_STRENGTH);
 
+const BOUNCE_FIRST_THRESHOLD = 4.0 / 11.0;
+const BOUNCE_SECOND_THRESHOLD = 8.0 / 11.0;
+const BOUNCE_THIRD_THRESHOLD = 9.0 / 10.0;
+
+// helper constants
+const ca = 4356.0 / 361.0;
+const cb = 35442.0 / 1805.0;
+const cc = 16061.0 / 1805.0;
+
+export const bounceOut = (p: number) => {
+  const p2 = p * p;
+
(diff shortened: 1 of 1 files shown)


magnusriga:
Thanks, @Andarist. Any idea when we can expect a merge?

magnusriga:
Also, seems it was previously implemented and then somehow dropped at some point:

https://github.com/Popmotion/popmotion/pull/11/files/c19b614f63458a07ae5266c4d43c9e90bc60c60a..80401f56f76ff26d098d63b6d23a4fcd6bbb2342#diff-2e624fdc150cdabb493037ebbd78aa4f

InventingWithMonster:
@magnusriga Yeah the reason I took it out was the idea that this kind of thing should be specifically handled by real physics. I do however understand the value of having it as part of a specific-duration animation.

I'll merge now but a publish might have to be Sunday/Monday

magnusriga:
Thank you, @InventingWithMonster , and thanks for a terrific library.

Andarist:
> Was you implementation based on the Penner function?

Sort of - yes.

InventingWithMonster:
Published as `@popmotion/easing@1.0.2`

magnusriga:
@InventingWithMonster

Thank you, Matt.

Would it be interesting if I implemented the other Penner function easings that are missing as well?

PR #746, opened 2019-02-28
Fix issue with leaving child being scheduled for removal on completes…

Andarist:
… non-exit poses


InventingWithMonster:
Ah great thanks for this!

InventingWithMonster:
Published as `react-pose@4.0.8`

PR #751, opened 2019-03-06
Use cross-realm safe "instanceof" checks

### LINKED ISSUE #723: [react-pose] Custom window for instanceof checks when running in an iframe
**Is your feature request related to a problem? Please describe.**

When rendering a posed element inside an IFrame using something like 
https://github.com/ryanseddon/react-frame-component the instanceof checks in pose fail, because they are instances of `Element` on the IFrame `contentWindow` rather than the browser global window.

**Describe the solution you'd like**
It be nicest if pose could determine the window context automatically, perhaps with `ownerWindow` or something...otherwise a way to specify a window context? I'm not quite sure where...since it feels out of place in pose config

Andarist:
fixes #723

Can't say I've tested this thoroughly, just got ball rolling 😉 

Could any of you @bradbarrow @quentez verify this?

### DIFF
--- a/packages/popmotion-pose/src/dom/flip.ts
+++ b/packages/popmotion-pose/src/dom/flip.ts
@@ -28,7 +28,8 @@ const hasPositionalProps = (pose: Pose) =>
   Object.keys(pose).some(checkPositionalProp);
 
 export const isFlipPose = (flip: boolean, key: string, state: PoserState) =>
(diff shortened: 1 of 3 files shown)


InventingWithMonster:
Looks like trouble 😄

Where are these checks failing currently?

Andarist:
> Where are these checks failing currently?

At all modified places 😉

PR #760, opened 2019-03-21
Improve framesync treeshakability

Andarist:
A simple line like this:
https://github.com/Andarist/use-smooth-scroll/blob/b1bc7ffe50495d32dd3251dbef56cef107b11812/src/index.js#L2
Is causing whole `framesync` to be imported because of this:
https://github.com/Popmotion/popmotion/blob/66ce2b87a727ec4d879a7013caf74bcec691d582/packages/popcorn/src/utils/smooth.ts#L2
Which in turn results in framesync eager initialization (its dynamic through reduce and such) and which causes WeakSets to be instantiated (they might be missing in older environments):
https://github.com/Popmotion/popmotion/blob/66ce2b87a727ec4d879a7013caf74bcec691d582/packages/framesync/src/create-render-step.ts#L15-L16

Long story short - this improves framesync treeshakeability if it stays unused. How? By adding pure annotations AND splitting steps/sync/cancelSync into 3 loops to avoid top-level property access which is treeshaking deopt (minifier cant assume getters to be pure).


packages/framesync/src/types.ts
```diff
@@ -15,13 +15,7 @@ export interface Step {
   process: (frame: FrameData) => void;
 }
 
-export enum StepId {
-  Read = 'read',
-  Update = 'update',
-  Render = 'render',
-  PostRender = 'postRender',
-  FixedUpdate = 'fixedUpdate'
```
Andarist: I've removed `fixedUpdate` entirely - it seems unused, couldnt find it anywhere.

Andarist:
@InventingWithMonster sorry, i've missed your reply

> Clever - would this still work if we replaced reduce with for loops?

Not quite, for loops would probably have to produce some side effects outside of them. The key to making things tree shakeable is often:
- containing local side effects within a statement
- ensuring that such things are contained within a simple~ single top level statement, in this case a single `reduce` call, because a single statement can be easily annotated as PURE

Andarist:
@InventingWithMonster could u take a look at this? I've rebased this so should be mergeable

PR #762, opened 2019-03-24
Unset event endPose when setting startPose

### LINKED ISSUE #761: Interia `dragEnd` after first drag - React Pose and Popmotion
## Ensure you're using the latest version
Tested on 4.0.8 and 2.0.0
## Add the package name to bug title
React-pose, popmotion interia
## Describe the bug
A Posed Div is created and it marked as "draggable" on the y axis. A dragEnd transition is setup and using popmotion Interia. 

Using posed 4.0.8 results in the first drag using interia when released. The second drag no longer uses interia. 

## How to reproduce
1. Go to https://codepen.io/anon/pen/XGoOKq?editors=0010
2. Drag the white "drawer" and release for interia. Do it again, notice it fails
3. Click on settings. Then JavaScript. Change the unpkg URL to use react-pose@2.0.0. Save and close
4. Notice the drag and release behavior works every time
## Expected behaviour

When dragging and releasing the "drawer" it will use interia motion every time.
## Link to CodePen example (or similar)
 Go to https://codepen.io/anon/pen/XGoOKq?editors=0010

## Device information
 - Device: [e.g. iPhone6] Macbook
 - OS: [e.g. iOS] OSX 10.14.3
 - Browser [e.g. chrome, safari] Firefox, Chrome 
 - Version [e.g. 22] Chrome 73, Firefox 66

## Additional context
N/A

Andarist:
fixes #761

However it's hard to tell if it doesn't break anything - it seems reasonable though to do this anyway. Otherwise endPose stay in the activePoses and when we unset startPose on 2nd interaction it gets set (incorrectly) in unset here:
https://github.com/Popmotion/popmotion/blob/6a4e0f2740b68a576a02a1c7f04f26c99542217d/packages/pose-core/src/index.ts#L142

Which caused endPose to be set twice when ending second interaction and messes up velocity and such

### DIFF
--- a/packages/popmotion-pose/src/dom/events.ts
+++ b/packages/popmotion-pose/src/dom/events.ts
@@ -56,6 +56,7 @@ const makeUIEventApplicator = ({
 
   const eventStartListener = listen(element, startEvents).start(
     (startEvent: MouseEvent | TouchEvent) => {
+      poser.unset(endPose);
       poser.set(startPose);
 
       if (startCallback && config[startCallback])


bringking:
@Andarist @InventingWithMonster any idea if/when this will land? We have a new feature that is dependent on this working and I would hate to try and roll back to 2.x

InventingWithMonster:
Yeah good fix! @bringking I'll release in the next 10 min

PR #763, opened 2019-03-24
Fix inertia action being completed in the middle of its internal sequ…

Andarist:
…ence

While reading the source I've noticed that inertia is being completed too soon - it sometimes runs a decay first and later chains a spring to it, but the action was completed unconditionally in the animation's complete callback so after mentioned decay and not after the whole sequence. This could have led to the latter spring interfering with any other actions chained to inertia.

I've not tested this thoroughly though, so please read the code carefully before merging.


InventingWithMonster:
Great catch, thanks for this

PR #775, opened 2019-04-11
Fix fatal error in chrome < 36

Deadly0:
Google chrome < 36 do not support css property `transform` without vendor prefix, so `transform` will be undefined and popmotion code will rise the error and crash whole react app:
`Uncaught TypeError: Cannot call method 'match' of undefined`

This PR prevent fatal error by skipping transition.


InventingWithMonster:
Great catch thanks

InventingWithMonster:
Published as `3.4.8`

Issue #498, opened 2018-09-18
Vue-Pose: onPoseComplete on exit?


thariddler:
I need to trigger function on Pose exit complete. Trying this, but it's triggered both: on enter and on exit. 

```js
Shade: posed.div({
      enter: {
        opacity: 1,
        scale: 1,
        beforeChildren: true,
        transition: { duration: 200, ease: 'linear' }
      },
      exit: {
        opacity: 0,
        scale: 0.5,
        afterChildren: true,
        transition: { duration: 200, ease: 'linear' },
        onPoseComplete: () => {
          console.log('complete')
        }
      }
    })
```

InventingWithMonster:
Hmm it’s weird it’s triggered at all to be honest, as onPoseComplete isn’t
a valid prop on the actual pose. It should be passed to the component
itself in the Vue template.

So you’d add it conditionally when you’re removing a component. I haven’t
actually tried this in Vue so I don’t know if it’ll update the components
props if you’re removing an element, I had to release a specific feature
for this in React but Vue works differently with leaving components

therddlr:
On component itself? Like this?

```js
Shade: posed.div({
      enter: {
        opacity: 1,
        scale: 1,
        beforeChildren: true,
        transition: { duration: 200, ease: 'linear' }
      },
      exit: {
        opacity: 0,
        scale: 0.5,
        afterChildren: true,
        transition: { duration: 200, ease: 'linear' }
      },
      onPoseComplete: () => {
        console.log('complete')
      }
    })
```
This way I have an error 

```
Error in mounted hook: "Error: HEY, LISTEN! Pose 'onPoseComplete' is of invalid type. All poses should be objects."
```

Not sure how to use it elsewhere?

InventingWithMonster:
Sorry like

<Shade :onPoseComplete=“yourMethod”

therddlr:
I made a reproduction https://codesandbox.io/s/mymv6ljxkj

therddlr:
Any tips why it's not working? Is it my code or it's a bug?

InventingWithMonster:
I'll need to take a look tomorrow morning. Might be a bug?

InventingWithMonster:
Ah yeah it's this line:

https://github.com/Popmotion/popmotion/blob/master/packages/vue-pose/src/components/pose-transition.ts#L23

[...]

PR #784, opened 2019-04-24
fix: the poseComplete event can not work #498

snowyu:
Compliance with the vue specification

* Change onPoseComplete prop to the @poseComplete vue event.

**Note:**

* put the poseComplete event on pose-transition now.

More detail to see #498 .


InventingWithMonster:
Excellent thanks!

InventingWithMonster:
Published as 0.5.1

Issue #590, opened 2018-10-26
[react native] text within nested posed component with animated position renders in low quality


kevinwolfcr:
## Version
`react-native-pose`: `^0.9.0`

## Describe the bug
Text is rendering in low quality when it is within a posed view with animated position and that view is within another posed view which also animates position.

## How to reproduce
1. Create a posed component with animated position and put a text inside it.
2. Wrap that component and animate the position of the wrapper component.
3. The text will now be in low quality.

## Expected behaviour
Text rendered in normal quality.

## Link to CodePen example (or similar)
https://snack.expo.io/@kevinwolf/nested-pop-issue

## Device information
In which environment are you reproducing the bug?
 - Device: Any iPhone
 - OS: iOS
 - Browser N/A
 - Version `^0.9.0`

### Component without wrapper component
![simulator screen shot - iphone 6 - 2018-10-25 at 21 12 52](https://user-images.githubusercontent.com/3157426/47542488-e5fe1f80-d89a-11e8-9340-f3d96bd3205c.png)

### Component wrapped
![simulator screen shot - iphone 6 - 2018-10-25 at 21 13 32](https://user-images.githubusercontent.com/3157426/47542494-eac2d380-d89a-11e8-8e70-b64cbe5fad9b.png)

kevinwolfcr:
Any update on this?

kevinwolfcr:
Didn't get a response, had to switch back to `LayoutAnimation` :/

juliusbangert:
I came across the same problem I asked for [help on stackoverflow](https://stackoverflow.com/questions/56264783/weird-antialias-artefacting-with-react-native-on-nested-posed-components)

Did you ever find a solution?

Issue #660, opened 2018-12-05
[react-pose] Scaled text is blurry in demos


amcdnl:
## Issue
On the demo page, if you look carefully at the interactive example when you hover you will notice the text is blurry.

![img](https://content.screencast.com/users/amcdaniel22/folders/Snagit/media/25fda2f3-9dd6-4b6e-b7ad-c8c7eafa62b1/2018-12-05_07-23-06.png)

This seems like a pretty common issue when doing transforms. Some people suggested setting the `perspective` property on the transform ( https://stackoverflow.com/questions/14677490/blurry-text-after-using-css-transform-scale-in-chrome ) but thats not possible given the component is overriding the transform.

Any suggestions/ideas?

InventingWithMonster:
I think perspective would still bring up this issue because the reason
these animations are performant is the element has already been rendered
and sent to the GPU as a raster image.

The only way around it (this applies to CSS too) is to render at a higher
resolution and scale down. In other words make the element larger to begin
with and then set scale: 0.9 (for instance) as the smaller scale and scale:
1 as the larger scale

amcdnl:
Gotcha - Thanks for the info - Closing.

amcdnl:
@InventingWithMonster - If you do `will-change: transform;` it removes the blur! :) - Might want to consider applying that automatically?

Aleksion:
@InventingWithMonster this also happens on react native. Performance is phenomenal when I use x or y to animate, but all text in the animated container becomes blurry as a result of it:

```typescript
{
	slideUp: (end: number = 0, start: number = -500) => ({
		[Animations.Shown]: {
			y: end,
			transition: {
				delay: 50,
				duration: 350,
				useNativeDriver: true,
			},
		},
		[Animations.Hidden]: {
			y: start,
			transition: {
				delay: 50,
				duration: 350,
				useNativeDriver: true,
			},
		},
	})
	}

```

twillstudio:
Hi guys, I am facing a similar issue on react-native-pose. i am using Transitions in order to slide booking components one by one. Here is a short part of my code for you to understand better what i'm doing.

```
<Transition
    enterBookingFlow={this.state.enterBookingFlow}
    deltaIVProduct={this.state.widthSlideIVProduct} 
    deltaLocation={this.state.widthSlideLocation} 
    deltaTime={this.state.widthSlideTime}
    >
<BoxListProduct/>
{
          this.props.dataFlow.currentPage == 'Time'?
[...]

Issue #404, opened 2018-07-23
low res render when nesting posed.View's in react-native


chrisdrackett:
Snack Here: https://snack.expo.io/@multiscale/pose-bug
(best to view on an actual device as the issue isn't clear on the in-page simulator)

when nesting two `posed.View`'s that animate translate (`x` and/or `y`) on react native the whole view will be rendered in a lower resolution. This happens both during and between animations.

therealravitej:
Facing the same issue.

chrisdrackett:
I tried to update my example to `0.8.1` but ran into an error when trying to use two pose elements together:

https://snack.expo.io/@multiscale/pose-bug-2

ChrisBews:
I'm facing the same issue here. I'm surprised there haven't been more complaints about this, as it pretty much means we can't use this library. If I had to guess, I'd say that it's rasterizing the components at their stated pixel dimensions, when it should actually be factoring the screen's pixel density. Further nesting just compounds the issue (the render quality of each component just gets worse the further you nest).

Would be good if this one was moved up the priority list - if React Native is considered an important target for the library.

ChrisBews:
Further to this - I've found the problem code. It's this line: https://github.com/Popmotion/popmotion/blob/a8ca84eea52c7055657cca15dd77b6480a309ac7/packages/react-native-pose/src/inc/utils.ts#L77

Is there any reason we're forcing perspective here? Without perspective set, the blurring goes away and everything looks great again. I'm sure there was a good reason as to why it was added, but it would be great if we could choose to override this value, or just prevent it getting added. Thanks!

InventingWithMonster:
Yeah I believe it was an Android issue, but due to my lack of comments it's
hard to remember what. That was a while ago though - if you open a PR with
this line removed I'll publish a version without it and deal with the
problems that do or don't arise as they come.

ChrisBews:
I've created that PR here - https://github.com/Popmotion/popmotion/pull/802

Thanks for being so responsive - your work on this library is invaluable, especially when working with React Native!

PR #802, opened 2019-05-17
Remove hard-coded perspective value when creating a transform properties

ChrisBews:
This was resulting in low quality rasterisation on iOS, which got increasingly worse the more deeply nested animations were.

Fixes: https://github.com/Popmotion/popmotion/issues/590, https://github.com/Popmotion/popmotion/issues/660 (the second half of this issue was this React Native issue), https://github.com/Popmotion/popmotion/issues/404

It's worth noting in this PR that @InventingWithMonster believes this may have originally been added to solve an Android issue, but isn't sure what that issue was, and I can't work it out either.

### DIFF
--- a/packages/react-native-pose/src/inc/utils.ts
+++ b/packages/react-native-pose/src/inc/utils.ts
@@ -74,7 +74,7 @@ export const getStylesFromPoser = (poser: AnimatedPoser) => {
         }
         return acc;
       },
-      [{ perspective: 1000 }] as TransformList
+      [] as TransformList
     );
   }
 


InventingWithMonster:
Thanks for this man - I'll publish and cross my fingers

InventingWithMonster:
Published as `react-native-pose@0.9.1`

ChrisBews:
Just updated and it's working great, thanks for your help!

I'll act as your human shield if a suspicious-looking Android problem suddenly rears it's head after this...

PR #823, opened 2019-07-05
Fix for timelines stagger delays

gaou-piou:


InventingWithMonster:
Thanks for this PR! I possibly won't be publishing before Popmotion 9 but it's good to get the logic fixed.

PR #827, opened 2019-07-17
Update html.md

stokesman:
I was trying to set scrollY and it wasn't working. So unless that's supposed to be aliased, just updating the docs would be nice


InventingWithMonster:
Thanks for this!

PR #836, opened 2019-10-01
Fixed unusable browser builds

### LINKED ISSUE #835: Stylefire: Broken "production" builds
## Describe the bug
Every version since 6.0.0 includes `process` without stripping it out of the final build, making it impossible to use any version on https://unpkg.com/stylefire

## How to reproduce
https://unpkg.com/stylefire@6.0.0/dist/stylefire.min.js or https://unpkg.com/stylefire@6.0.0/dist/stylefire.js

## Expected behaviour
`process` should not be in the final build particularly if this is not es modules.

## Device information

Akkuma:
Resolves #835 by replacing `process.env.NODE_ENV` with the appropriate `production`.

### DIFF
--- a/packages/stylefire/CHANGELOG.md
+++ b/packages/stylefire/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 Stylefire adheres to [Semantic Versioning](http://semver.org/).
 
+## [6.0.12] 2019-10-01
+
+### Fixed
+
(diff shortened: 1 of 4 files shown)


InventingWithMonster:
Nice one thanks for this!

InventingWithMonster:
Published as 6.0.12

PR #837, opened 2019-10-01
Updated to latest stylefire

Akkuma:
Stylefire is fairly out of date, which in my case was causing significant problems with SVG animations. This updates it so that people aren't running outdated stylefire's in their popmotion pure.

### DIFF
--- a/packages/popmotion/CHANGELOG.md
+++ b/packages/popmotion/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 Popmotion adheres to [Semantic Versioning](http://semver.org/).
 
+## [8.7.1] 2019-10-01
+
+### Update
+
+- `stylefire@6.0.11`
+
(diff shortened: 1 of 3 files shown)


InventingWithMonster:
There's been some changes to the Stylefire API that mean this would actually be a major version increase. I'm actually going to remove Stylefire in the next major and you can also consider it a seperate dependency here by installing the latest `stylefire` and importing directly from that package.

Akkuma:
> Thanks for the PR but there's been some changes to the Stylefire API that mean this would actually be a major version increase. I'm actually going to remove Stylefire in the next major and you can also consider it a seperate dependency here by installing the latest `stylefire` and importing directly from that package.

I'd highly recommend at least changing the documentation everywhere to reflect that popmotion's internal stylefire should not be used at all due to being severely outdated with guaranteed bugs.

PR #838, opened 2019-10-02
Added pause/resume to chain and delay

Akkuma:
I ran into the need to be able to pause a `stagger` and it seemed strange that you could in theory pause tweens in a `parallel`, but couldn't pause a `stagger` which uses a `parallel` because it relied upon `chain` & `delay`.

- `chain` passes through `pause` and `resume`, much like  it passes it through for `stop`.
- `delay` will `pause` by `clearTimeout` and taking the remaining time to create a new timeout if there is any, which `resume` will use to create the new remaining time.

### DIFF
--- a/packages/popmotion/CHANGELOG.md
+++ b/packages/popmotion/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 Popmotion adheres to [Semantic Versioning](http://semver.org/).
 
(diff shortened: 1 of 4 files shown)


InventingWithMonster:
Closing this as I'm overhauling code related to animation playback.

PR #839, opened 2019-10-08
Standardising x/y across HTML and SVG

InventingWithMonster:


amcdnl:
This is breaking this: https://github.com/framer/motion/issues/384

Issue #842, opened 2019-10-30
[feature request] pure release with the dist folder only includes single index.d.ts index.es.js index.js


morlay:
**Is your feature request related to a problem? Please describe.**
There are some unnecessary files in npm pkg
![image](https://user-images.githubusercontent.com/1667873/67834401-47964100-fb22-11e9-9b56-8f4836f85d5c.png)

**Describe the solution you'd like**
we could publish module with the `dist` folder only.

**Describe alternatives you've considered**
`index.js` and `index.es.js` are already built by rollup. 
with these plugin [rollup-plugin-dts](https://github.com/Swatinem/rollup-plugin-dts), we could built single `index.d.ts` too (from the lib folder built by typescript).

**Additional context**
final release could be
![image](https://user-images.githubusercontent.com/1667873/67834486-7a403980-fb22-11e9-8e82-a6526d712114.png)

PR #843, opened 2019-10-30
example for #842

morlay:
results like:
image

block by a small issue https://github.com/Swatinem/rollup-plugin-dts/pull/65


packages/popmotion/src/reactions/value.ts
```diff
@@ -79,7 +79,7 @@ export class ValueReaction extends BaseMulticast<ValueReaction> {
     sync.postRender(this.scheduleVelocityCheck);
   }
 
-  scheduleVelocityCheck = () => sync.postRender(this.velocityCheck);
+  scheduleVelocityCheck = (): Process => sync.postRender(this.velocityCheck);
```
morlay: just to fix typescript declaration generator

PR #854, opened 2019-12-05
fix color test and transform in style-value-types package

nathanredblur:
I'm fixing several color validations for missing cases like `color.test({ red: '' }) === false` or transforms like `hsla.transform({ lightness: 45, alpha: 1 }) === 'hsla(0, 0%, 45%, 1)'`. All cases where data is missing or incorrectly assumed like a valid color input value.

### DIFF
--- a/packages/style-value-types/src/_tests/index.test.ts
+++ b/packages/style-value-types/src/_tests/index.test.ts
@@ -127,6 +127,9 @@ describe('rgba()', () => {
   it('should correctly test for colors', () => {
     expect(rgba.test('rgba(255, 0, 0, 0.5)')).toEqual(true);
     expect(rgba.test('rgba(255, 0, 0, 0.5) 0px')).toEqual(false);
+    expect(rgba.test(red)).toEqual(true);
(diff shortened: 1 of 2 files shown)


nathanredblur:
CircleCI build has issues not related with my changes.

InventingWithMonster:
Thanks for the PR but the tests are designed to be performant and practical rather than exhaustive. I'm going to add a comment to clarify this.

nathanredblur:
is not only test, is also issues with the library identifying  like a valid rgb color an element that is not.

please check the changes in this file.
https://github.com/Popmotion/popmotion/pull/854/files#diff-b9c4a35a5bd787070e5e8a535ac682d1

you determine like a valid rgb an object only because this has a red color instead of the other colors too, this derive in issues on try to use or transform that color later.

this is for good.

InventingWithMonster:
Type systems will prevent 

```
color.test({ hue: 0, red: 0 })
color.test({ red: '' })
```

Your changes are more correct and thorough but they are also more inefficient. The code is written for simple practical purposes, for instance quickly determining whether it's a color or which color type something is.

PR #874, opened 2020-04-27
Fix decay and inertia, plus refinements

stokesman:
This PR includes the fixes from #873 plus some changes that lighten and tighten inertia's implementation a bit.

[4c9f561 streamline subordinate action creation in inertia](https://github.com/Popmotion/popmotion/commit/4c9f561d672caf9855f816f39f8645423a7c1547)
`inertia` exports a `VectorActionFactory` and imports `decay` and `spring` as such. Thus when inertia creates the decay or spring animations redundant logic for vector actions is ran. By importing and using the single versions of decay and spring that is avoided. 

[eb254d4 tighten logic for switching animations in inertia](https://github.com/Popmotion/popmotion/commit/eb254d4ddb7bc0c2f76f1ede054396b9838b0082)
I couldn't help but wonder why inertia didn't use a while predicate on decay for the logic of checking boundaries and switching to spring. So it tried it out and found a few other small ways to refine the implementation.

[b07fdef remove value reaction from inertia](https://github.com/Popmotion/popmotion/commit/b07fdef021540cd1e103e2d34fe8273672c3958c)
It seemed like manually tracking velocity to drop reliance on value reaction was worthwhile for slightly less overhead. Also in the odd case that someone just wants inertia and not reactions they can be treeshakened.

### DIFF
--- a/packages/popmotion/CHANGELOG.md
+++ b/packages/popmotion/CHANGELOG.md
@@ -2,6 +2,18 @@
 
 Popmotion adheres to [Semantic Versioning](http://semver.org/).
 
+## [8.7.2] 2020-04-28
+
+### Fixed
+
+- Decay and inertia correctly start motion at from prop when using modifyTarget
+- Decay and inertia no longer round target calculation
+- Inertia calls modifyTarget even with zero-velocity
+
+### Changed
+
+- Inertia implementation refinements
+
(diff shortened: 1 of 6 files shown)


InventingWithMonster:
These are great changes, thanks for taking the time to look at this!

PR #876, opened 2020-04-30
mend rollup config generator

### LINKED ISSUE #875: Unbundled dependencies in UMD builds (Stylefire published that way since 7.0.2)
The generator for Rollup configs is missing the resolve plugin. There are also some other things that were lost in the move to the generator.

stokesman:
getting back details lost when switching to generator
- import and use resolve plugin for umd/umdProd
- use no declaration typescript config for all but cjs
- use unpkg field when available for file in umdProd
- remove doubled up yarn build in root scripts build

fixes #875

I've not tested every package but seems to work well for pure, stylefire, framesync and popcorn.

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,7 +5,7 @@
   "homepage": "https://popmotion.io",
   "scripts": {
     "bootstrap": "lerna bootstrap",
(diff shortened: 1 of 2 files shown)


InventingWithMonster:
Excellent thanks for this!

PR #879, opened 2020-05-11
Indicate that package has no sideEffects

voxpelli:
To help tree shaking in Webpack as Webpack for compatibility reasons otherwise will assume that we package will have sideEffects.

See: https://github.com/stereobooster/package.json#sideeffects
And: https://webpack.js.org/guides/tree-shaking/


stokesman:
It does have side effects. Tree-shaking can still happen since pure functions are annotated. I think Webpack just needs to be in [production mode](https://webpack.js.org/configuration/mode/#mode-production).

PR #890, opened 2020-08-27
Fix stylefire docs

dminkovsky:
Looks like these props are now 'scrollTop' and 'scrollLeft'. 

--- 

A slightly related question: 

Since stylefire is no longer exported from the popmotion package, is there some other way you are expected to actually execute an animation with v9?

Or are you still supposed to create/user a styler like before, but are now expected to `import styler from 'stylefire'`?

Thanks


dminkovsky:
Hmm looks like they're correct here https://github.com/Popmotion/popmotion/blob/9/packages/stylefire/docs/api/stylers/html.md.

PR #944, opened 2022-05-23
Update `export` maps for TypeScript 4.7+

rossng:
TypeScript 4.7 (soon to be released) now has support for `exports` maps in `package.json` and ES module output. In packages that already provide an `exports` field, this can end up hiding the `types` field, causing builds to break.

These changes should resolve that by specifying `types` in the `exports` map itself.

### DIFF
--- a/packages/framesync/package.json
+++ b/packages/framesync/package.json
@@ -12,6 +12,7 @@
     "unpkg": "./dist/framesync.min.js",
     "exports": {
         ".": {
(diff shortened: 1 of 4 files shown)


rossng:
@mattgperry would it be possible to merge this?

mattgperry (approved):

mattgperry:
Looks great - thanks!

